PR #9, opened 2025-10-06
fix: add version command and arg

### LINKED ISSUE #8: version command for the tool
👋 Would be good to add a version command for the tool, like `container-compose version`

bassco:
Closes #8 

Adds a version command to the tool.

It would need to be manually updated in Application.swift before release.

### DIFF
--- a/Sources/Container-Compose/Application.swift
+++ b/Sources/Container-Compose/Application.swift
@@ -12,7 +12,7 @@ import ArgumentParser
 struct Main: AsyncParsableCommand {
     private static let commandName: String = "container-compose"
     private static let version: String = "v0.5.1"
-    public static var versionString: String {
+    static var versionString: String {
         "\(commandName) version \(version)"
     }
     static let configuration: CommandConfiguration = .init(
--- a/Sources/Container-Compose/Commands/Version.swift
+++ b/Sources/Container-Compose/Commands/Version.swift
@@ -24,15 +24,14 @@
 import ArgumentParser
 import Foundation
 
-internal struct Version: ParsableCommand {
-    public init() {}
+struct Version: ParsableCommand {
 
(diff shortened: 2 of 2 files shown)


Mcrich23:
I wonder if it would be better to do `container-compose --version` instead of `container-compose version`

bassco:
`--version` is free from the Framework.
The `version` command is how upstream docker exposes the versions. Although they have the same value for both outputs.
 
```shell
 $ ./.build/arm64-apple-macosx/release/Container-Compose version
Container Compose version v0.5.1
$ ./.build/arm64-apple-macosx/release/Container-Compose --version
v0.5.1

$ docker compose version
Docker Compose version v2.23.0-desktop.1
$ docker compose --version
Docker Compose version v2.23.0-desktop.1
```

I've added a commit to allow for the output to be the same across both invocations.

Mcrich23:
Sounds good!

Mcrich23 (approved):

bassco:
> No both are great. I just want to keep tight access controls. There isn't a reason for the Version command to be accessible outside the module, so we don't need to make it public. Same for the commandName and version variables, but in that case they should be marked as private.

Setting the variables to be private - good call.
Version command changed to internal scope.

Mcrich23:
Thank you so much for your effort. Just a few small nitpicks and then this can get merged.

Mcrich23 (changes requested):
Somehow didn't finish the review before, so i'm just doing that

bassco:
> Somehow didn't finish the review before, so i'm just doing that
yeah, the github ui for the "double-submit" for review comments trips me up from time to time too. 
Thanks for the scoping insights.

Mcrich23:
Thank you for taking the time to scope it well.

Mcrich23 (approved):

PR #31, opened 2025-11-25
fixed checks for running containers during "up" command

TomasLudvik:
Currently all checks failed on timeout even when containers started correctly

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeUp.swift
+++ b/Sources/Container-Compose/Commands/ComposeUp.swift
@@ -180,11 +180,7 @@ public struct ComposeUp: AsyncParsableCommand, @unchecked Sendable {
         fatalError("unreachable")
     }
 
-    private func getIPForRunningService(_ serviceName: String) async throws -> String? {
-        guard let projectName else { return nil }
-
-        let containerName = "\(projectName)-\(serviceName)"
-
+    private func getIPForContainer(_ containerName: String) async throws -> String? {
         let container = try await ClientContainer.get(id: containerName)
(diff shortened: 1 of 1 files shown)


Mcrich23:
Sorry, I'm a bit confused. What's the exact issue you are trying to solve?

Also, in the future, please open an issue and then make a pull request referencing the issue. This will make issue tracking easier.

Cyb3rDudu:
Just hit all three problems this PR addresses while migrating a real Docker stack:

1. **Wait-timeout false-alarm**: `up -d` prints `Error Domain=ContainerWait Code=1 "Timed out waiting for container..."` even when the container actually started. Looking at the wait loop on `main`, the timeout is hit because `try? await client.get()` returns nil briefly while the container is being created — there's no retry and no distinction between "doesn't exist yet" and "exists but not running." This PR's `do/catch` keeps polling through the not-found phase, which is what's wanted.

2. **Container-already-exists**: After a failed run, the container often lingers in stopped state. On `main`, the next `up` blows up with `Error: container with id <name> already exists`. This PR's early-return when an existing container is already running, plus the explicit "exists with status: X" message, makes recovery much smoother. (Related: #38 covers the cleanup side; this is the recovery side.)

3. **IP extraction**: subtle but real — `main` reads `$0.ipv4Gateway.description`, which is the network gateway, not the container IP. This PR uses `CIDRAddress($0.address).address.description` which gets the container's own IP. That matters for any service-IP env var injection.

Conflict against current `main` is real — main refactored from `ClientContainer.get` static to `ContainerClient()` instance, plus reworked `composeDirectory` resolution.

PR #35, opened 2025-12-01
(More) complete support for depends_on

rtoohil:
My compose file has a more complex use of the `depends_on` functionality than Container-Compose used. 

I will admit that Swift isn't my best language, so I worked with Claude Code to build this PR. This adds support for various `depends_on` syntaxes such as:

```
1. Simple string (no colon after service name):

  depends_on: mysql

  2. Simple array:

  depends_on:
    - mysql
    - redis

  3. Dictionary without conditions (empty/null values):

  depends_on:
    mysql:
    redis:

  4. Dictionary with conditions (full format):

  depends_on:
    mysql:
      condition: service_healthy
    redis:
      condition: service_started
```

It doesn't handle the health check well yet, but we can expand upon that. This handled my more complex `depends_on` case.

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Service.swift
+++ b/Sources/Container-Compose/Codable Structs/Service.swift
@@ -185,9 +185,7 @@ public struct Service: Codable, Hashable {
         if let cmdArray = try? container.decodeIfPresent([String].self, forKey: .command) {
             command = cmdArray
         } else if let cmdString = try? container.decodeIfPresent(String.self, forKey: .command) {
-            // When command is a string, wrap it in a shell to properly handle arguments
-            // This matches Docker Compose behavior for string commands
-            command = ["/bin/sh", "-c", cmdString]
+            command = [cmdString]
         } else {
             command = nil
         }


Mcrich23:
This is true. I will do a manual review of your PR when I have a chance. One way that would be incredibly helpful to contribute if you would like would be to document the current status of compose yaml feature support. 

Feel free not to if you don't want to or don't have the capacity to.

Mcrich23:
Thank you. Can you please add tests for this? I want to ensure in the future that this kind of compose file will continue to build.

rtoohil:
They are all passing on main, which is when I started to investigate, but the test "Test compose with complex dependency chain” returns success even though a couple of the containers never start due to errors.

When updating the depends_on logic, we catch this error now, so the test fails. I’m partially through debugging why the containers are failing (one was the wrong argument passed into it — which is now working, but still hunting it down).

If you see different behavior (i.e. the containers for app and wordpress do start properly), let me know. But I think I’ll make progress on fixing that issue in the next day or so.

rtoohil:
Sorry for the long delay (real life). This has the same tests passing as on main (the Real World tests fail on main on my machine), but I reverted the change that broke the other test. That test is now passing.

I do think that there's some command handling that may need to change to get these other tests working, but that can be a separate PR.

Issue #33, opened 2025-11-29
The presence of a docker-compose.yml file overrides the -f argument


rtoohil:
(First, super excited about this project -- I'd really like to see how `container` compares to `docker` with one of my containerized apps.)

When you run `container-compose up -f <foo>` if there's a docker-compose.yml file present, it defaults to that rather than using the file passed via the `-f` argument.

It looks to me (as a novice, so I could be quite wrong) that the check on lines 85-90 of `ComposeUp.swift` finds one of the known files and then sets the filename before it then reads the path provided.

ryan106:
Looks like this was fixed by #63 — I just tested against current `main`:

```
$ container-compose -f other.yaml up -d
```

with a `docker-compose.yml` decoy also present in the cwd, and it correctly used `other.yaml`. `composePath` in `ComposeUp.swift`/`ComposeBuild.swift`/`ComposeDown.swift` all check `composeFileOptions.composeFilename` (the `-f` value) before the auto-discovery loop over `compose.yml`/`docker-compose.yml`/etc. Might be safe to close as a duplicate of #63 unless I'm missing a repro case that still breaks.

PR #36, opened 2025-12-01
Properly handle the -f flag to be the primary file when used by up or down

rtoohil:
As mentioned in #33, if you have one of the "key" compose file names, `up` and `down` will choose that file, even if you've provided a file via the `-f` flag. This changes the logic to choose and handle the `-f` flag appropriately.

With Swift not being my primary language, I've worked with Claude Code to build this PR.


Mcrich23 (changes requested):

Cyb3rDudu:
closing — superseded by #98 + composeFilename precedence (-f is already the primary file).

Issue #34, opened 2025-11-29
Support for Docker Compose `build context`


rtoohil:
Some of my compose files uses build context, i.e.

```
app:
  build:
    context: ./backend
```

Which tells `compose` to look in `./backend` for the appropriate files, including `Dockerfile`. It looks like `Build.swift` tries to support this, but it hasn't worked in my testing. In my app, there's a `Dockerfile` in `./backend`, but `compose up` doesn't find it.

Specifying a `Dockerfile` like:

```
 app:
    build: 
      context: ./backend
      dockerfile: ./backend/Dockerfile
```

also fails with this error: `Error: /Users/rtoohil/Development/stash is not a child of /Users/rtoohil/Development/stash/./backend`

Mcrich23:
Build context handling has been updated since this report. With `context: ./backend`, `dockerfile` should be relative to that directory, so `Dockerfile` would point to `./backend/Dockerfile`. The first example should now work without specifying it. Can you confirm whether it works with the current version?

PR #37, opened 2025-12-01
Adding support for `build: context`

rtoohil:
As mentioned in #34, only a small portion of the `build` syntax was supported, which didn't handle the `context` and `dockerfile` (and other options). This PR handles the additional capabilities (which has enabled my default compose file to work).

Swift is not my primary language, so I was assisted in building with Claude Code.

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeUp.swift
+++ b/Sources/Container-Compose/Commands/ComposeUp.swift
@@ -598,16 +598,25 @@ public struct ComposeUp: AsyncParsableCommand, @unchecked Sendable {
             return imageToRun
         }
 
+        // Resolve context path (handle absolute paths)
+        let contextPath: String
+        if buildConfig.context.hasPrefix("/") || buildConfig.context.hasPrefix("~") {
+            contextPath = buildConfig.context
+        } else {
+            contextPath = "\(self.cwd)/\(buildConfig.context)"
+        }
+
         // Build command arguments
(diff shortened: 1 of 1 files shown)


Mcrich23 (changes requested):

Cyb3rDudu:
Tested on current `main`. The context-resolution half of this PR is now redundant — main has `composeDirectory` and resolves `build.context` relative to the compose file. But the **dockerfile** path is still wrong: main builds it relative to `composeDirectory`, not relative to the resolved context.

```
$ tree /tmp/ctx-test
/tmp/ctx-test
├── compose-dir
│   └── test.yml      # services.ctxtest.build.context: ../sub
└── sub
    └── Dockerfile
```

```yaml
# /tmp/ctx-test/compose-dir/test.yml
services:
  ctxtest:
    build:
      context: ../sub
    command: ["cat", "/marker"]
```

```
$ container-compose build -f /tmp/ctx-test/compose-dir/test.yml
Error: dockerfile does not exist /tmp/ctx-test/compose-dir/Dockerfile
```

Per Compose spec, `dockerfile:` is relative to the resolved context, not the compose file. So the dockerfile half of this PR (`"\(contextPath)/\(buildConfig.dockerfile ?? "Dockerfile")"`) is the right semantic, just needs to use `composeDirectory`-resolved `contextPath` rather than cwd-based.

Mcrich23:
I agree and have a feeling we can get your PR merged a bit faster on this, so I will close this one for now.

Issue #38, opened 2025-12-01
compose down doesn't support named containers (or custom named containers)


rtoohil:
If you use `container_name: ` with a custom name in your compose file, `compose up` properly handles that, but `compose down` does not, and defaults to its internal naming, will fail to stop the container, and will not throw an error, because it eats the error.

PR incoming.

malclear:
I'm seeing `waitUntilServiceIsRunning` throw a "Timed out waiting for container" when running `up` and a custom container name is specified. The method, and two others, reconstructs the container name from `projectName` and `serviceName`. It does not respect the configured container name.
Also, the "Parameters" description in the function's comments is not accurate.

Mcrich23:
Custom names are now included in the shared container resolution, and the earlier startup and shutdown fixes should address this report. Can you try the original example against the current version if you're still using this setup? Interpolated container names are separate work in #102.

PR #39, opened 2025-12-01
Handle compose down with container_name case

rtoohil:
If you use container_name:  with a custom name in your compose file, compose up properly handles that, but compose down does not, and defaults to its internal naming, will fail to stop the container, and will not throw an error, because it eats the error.

See #38 

This should resolve that case. PR contributed to with Claude.

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeDown.swift
+++ b/Sources/Container-Compose/Commands/ComposeDown.swift
@@ -103,25 +103,37 @@ public struct ComposeDown: AsyncParsableCommand {
             })
         }
 
-        try await stopOldStuff(services.map({ $0.serviceName }), remove: false)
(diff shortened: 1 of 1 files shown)


lepus (approved):

Mcrich23 (changes requested):
This looks great! Thank you

Would it be possible to add a new dynamic test for this? You can just make a new file in `Tests/Container-Compose-DynamicTests` called `ComposeDownTests.swift`. It could probably just bring up a container and then shut it down and make sure that everything goes up and down.

rtoohil:
Thanks! I've been hit with the real life bug and haven't had time to get back to these PRs, but I appreciate the help of adding the test!

Mcrich23:
Merged this with #50.

PR #42, opened 2025-12-04
fix: support relative paths in host path volumes

### LINKED ISSUE #4: Support Relative Bind Mounts
It would be nice if this plugin supported the ability to specify relative bind mounts:

```yaml
services:
  app:
    image: node:latest
    ports:
      - 3000:3000
    volumes:
      - ./app:/app
```

Current output:

```
Info: No 'name' field found in docker-compose.yml. Using directory name as project name: test
Stopping container: test-app

--- Processing Services ---
["app"]
Pulling Image node:latest...
Info: Created missing host directory for volume: /Users/adrum/Developer/test/app/./app 
Note: Service 'app' is not explicitly connected to any networks. It will likely use the default bridge network.

Starting service: app
Starting app
----------------------------------------

app: Error: invalidArgument: "Invalid volume name './app': must match ^[A-Za-z0-9][A-Za-z0-9_.-]*$"

Error Domain=ContainerWait Code=1 "Timed out waiting for container 'test-app' to be running." UserInfo={NSLocalizedDescription=Timed out waiting for container 'test-app' to be running.}
```

ttys3:
- Convert relative paths (e.g., ./data/postgres) to normalized absolute paths
- Use URL.standardizedFileURL to remove ./ and ../ components
- Pass absolute paths to container run -v command instead of relative paths
- Fixes issue where container tool incorrectly treats relative paths as named volume names

The container tool requires absolute paths for bind mounts. Previously, relative paths like './data/postgres' were passed directly, causing the tool to validate them as named volume names (which must match ^[A-Za-z0-9][A-Za-z0-9_.-]*$), resulting in validation errors.

fix https://github.com/Mcrich23/Container-Compose/issues/4

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeUp.swift
+++ b/Sources/Container-Compose/Commands/ComposeUp.swift
@@ -664,13 +664,15 @@ public struct ComposeUp: AsyncParsableCommand, @unchecked Sendable {
             var isDirectory: ObjCBool = false
             // Ensure the path is absolute or relative to the current directory for FileManager
             let fullHostPath = (source.starts(with: "/") || source.starts(with: "~")) ? source : (cwd + "/" + source)
+            // Normalize the path to remove ./ and ../ components, and convert to absolute path
+            let normalizedHostPath = URL(fileURLWithPath: fullHostPath).standardizedFileURL.path(percentEncoded: false)
 
-            if fileManager.fileExists(atPath: fullHostPath, isDirectory: &isDirectory) {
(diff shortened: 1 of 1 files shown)


Mcrich23:
@ttys3 Unit testing should not take more than 10-15 minutes to add if you don't mind.

Cyb3rDudu:
@ttys3 — picked up your approach in #84 with the unit tests Morris asked for here. Same runtime fix (`URL.standardizedFileURL` to normalize and pass the absolute path to `-v`), just refactored into a static helper so it's directly testable. Credit to you in the PR body.

PR #50, opened 2026-01-30
feat: Providing tests for `ComposeDown`

oxisto:
This PR adds a dynamic tests for `ComposeDown` and incorporates the fix of #39 to correctly handle container_name.

### DIFF
--- a/Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
+++ b/Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
@@ -44,7 +44,7 @@ struct ComposeDownTests {
             containers.count == 2,
             "Expected 2 containers for \(project.name), found \(containers.count)")
 
-        #expect(containers.filter({ $0.status == .running }).count == 2, "Expected 2 running containers for \(project.name), found \(containers.count)")
+        #expect(containers.filter({ $0.status == .running }).count == 2, "Expected 2 running containers for \(project.name), found \(containers.filter({ $0.status == .running }).count)")
 
         var composeDown = try ComposeDown.parse(["--cwd", project.base.path(percentEncoded: false)])
         try await composeDown.run()
@@ -58,7 +58,7 @@ struct ComposeDownTests {
             containers.count == 2,
             "Expected 2 containers for \(project.name), found \(containers.count)")
 
-        #expect(containers.filter({ $0.status == .stopped}).count == 2, "Expected 2 stopped containers for \(project.name), found \(containers.count)")
+        #expect(containers.filter({ $0.status == .stopped}).count == 2, "Expected 2 stopped containers for \(project.name), found \(containers.filter({ $0.status == .stopped }).count)")
     }
 
     @Test("What goes up must come down - container_name")


Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        containers = try await ClientContainer.list()
+            .filter({
+                $0.configuration.id.contains(project.name)
+            })
+
+        #expect(
+            containers.count == 2,
+            "Expected 2 containers for \(project.name), found \(containers.count)")
+        #expect(
+            containers[0].status == .stopped,
```
Mcrich23: Could you please make this check that all of these containers are stopped instead of just the first one?

Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        var containers = try await ClientContainer.list()
+            .filter({
+                $0.configuration.id.contains(containerName)
+            })
+
+        #expect(
+            containers.count == 1,
+            "Expected 1 container with the name \(containerName), found \(containers.count)")
+        #expect(
+            containers[0].status == .running,
```
Mcrich23: Could you please make this check that all of these containers are running instead of just the first one?

Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        containers = try await ClientContainer.list()
+            .filter({
+                $0.configuration.id.contains(containerName)
+            })
+
+        #expect(
+            containers.count == 1,
+            "Expected 1 container with the name \(containerName), found \(containers.count)")
+        #expect(
+            containers[0].status == .stopped,
```
Mcrich23: Could you please make this check that all of these containers are stopped instead of just the first one?

Mcrich23 (changes requested):
After taking a deeper look, it mostly feels good, but I have some specific requests regarding tests to make sure our tests are robust and strong to bring into the future of Container-Compose.

Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
```diff
@@ -0,0 +1,112 @@
...
+
+import ContainerAPIClient
+import ContainerCommands
+import Foundation
+import TestHelpers
+import Testing
+
+@testable import ContainerComposeCore
+
+@Suite("Compose Down Tests")
```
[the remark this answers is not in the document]
oxisto: Done

Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
oxisto: There should only be one for this particular test-case :)

Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
```diff
@@ -0,0 +1,112 @@
...
+        var containers = try await ClientContainer.list()
+            .filter({
+                $0.configuration.id.contains(project.name)
+            })
+
+        #expect(
+            containers.count == 2,
+            "Expected 2 containers for \(project.name), found \(containers.count)")
+        #expect(
+            containers[0].status == .running,
```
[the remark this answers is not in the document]
Mcrich23: Interesting. Try manually restarting the container subsystem

Tests/Container-Compose-DynamicTests/ComposeDownTests.swift
Mcrich23: Maybe just do a filter for all that have a status of stopped and check that the count of that array is 0

Something like:
```
containers.filter { $0.status == .stopped }.count == 1
```

Mcrich23:
Lets see if we can fix the wordpress test for right now, and then we can fix it in another PR.

PR #60, opened 2026-03-06
fix: handle port mappings with explicit IP bindings

### LINKED ISSUE #59: Port mapping fails when compose file specifies explicit IP binding
## Bug Description

When using docker-compose.yml with port mappings that include explicit IP bindings (e.g., `127.0.0.1:5432:5432`), container-compose fails with:

```
Error: invalid publish IPv4 address: 0.0.0.0:127.0.0.1:5432:5432
```

## Root Cause

The port handling code in `ComposeUp.swift` blindly prepends `0.0.0.0:` to every port specification, without checking if the spec already contains an IP binding. This transforms `127.0.0.1:5432:5432` into the invalid `0.0.0.0:127.0.0.1:5432:5432` (4 colon-separated components instead of 3).

## Solution

Implemented a proper port parser `composePortToRunArg()` that:
- Recognizes various port formats: `PORT`, `HOST:PORT`, `IP:HOST:PORT`
- Preserves explicit IP bindings (IPv4, IPv6)
- Maintains protocol suffixes (/tcp, /udp)
- Only prepends `0.0.0.0:` when host part is just a port number

## Changes

- Added `composePortToRunArg()` helper in Helper Functions.swift
- Updated `configService()` to use the helper
- Added 8 comprehensive unit tests covering all formats

## Testing

All existing tests pass, and new tests verify correct handling of:
- Simple container port: `3000` → `0.0.0.0:3000:3000`
- Host:container same/different ports
- Explicit IPv4 binding: `127.0.0.1:5432:5432` (preserved)
- Explicit IPv6 binding: `[::1]:3000:3000` (preserved)
- Protocol suffixes: `3000:3000/tcp`
- Explicit IP with protocol: `127.0.0.1:5432:5432/tcp`

This fix resolves the invalid address error and makes container-compose compatible with standard Docker Compose port specifications that use explicit IP bindings (common for localhost-only services or specific interface binding).

ravshansbox:
## Summary

Fixes the "invalid publish IPv4 address" error when docker-compose.yml port mappings include explicit IP bindings (e.g., `127.0.0.1:5432:5432`).

## Changes

- **Added** `composePortToRunArg()` helper function in `Helper Functions.swift` that correctly parses Docker Compose port specifications into `container run -p` format
- **Updated** `configService()` in `ComposeUp.swift` to use the new helper instead of blindly prepending `0.0.0.0:`
- **Added** 8 unit tests in `HelperFunctionsTests.swift` covering all port format scenarios

## What was wrong

Previous code did:
```swift
runCommandArgs.append("0.0.0.0:\(resolvedPort)")
```
This produced invalid address `0.0.0.0:127.0.0.1:5432:5432` when the port spec already included an IP.

## How it works now

The new parser intelligently handles all formats:
- `3000` → `0.0.0.0:3000:3000` (auto-completes)
- `8080:3000` → `0.0.0.0:8080:3000` (adds IP binding)
- `127.0.0.1:5432:5432` → `127.0.0.1:5432:5432` (preserves explicit IP)
- `[::1]:3000:3000` → `[::1]:3000:3000` (preserves IPv6)
- All with optional `/tcp` or `/udp` suffixes preserved

## Testing

✅ All 70+ existing tests pass
✅ 8 new port conversion tests pass
✅ Build succeeds with no warnings

Closes #59

### DIFF
--- a/Tests/Container-Compose-DynamicTests/ComposeUpTests.swift
+++ b/Tests/Container-Compose-DynamicTests/ComposeUpTests.swift
@@ -267,6 +267,46 @@ struct ComposeUpTests {
                 #expect(appContainer.configuration.resources.cpus == 1)
                 #expect(appContainer.configuration.resources.memoryInBytes == 512.mib())
         }
+
+    @Test("Test compose up with explicit IP port mapping")
+    func testComposeUpWithExplicitIPPortMapping() async throws {
+        let yaml = """
+            version: "3.8"
+            services:
+                web:
+                    image: nginx:alpine
+                    ports:
+                        - "127.0.0.1:18081:80"
+            """
+
+        let project = try DockerComposeYamlFiles.copyYamlToTemporaryLocation(yaml: yaml)
+
+        var composeUp = try ComposeUp.parse(["-d", "--cwd", project.base.path(percentEncoded: false)])
+        try await composeUp.run()
+
+        var containers = try await ClientContainer.list()
+            .filter({
+                $0.configuration.id.contains(project.name)
+            })
+
+        guard let webContainer = containers.first(where: { $0.configuration.id == "\(project.name)-web" }) else {
+            throw Errors.containerNotFound
+        }
+
+        #expect(webContainer.status == .running)
+        #expect(webContainer.configuration.publishedPorts.map({ "\($0.hostAddress):\($0.hostPort):\($0.containerPort)" }) == ["127.0.0.1:18081:80"])
+
+        var composeDown = try ComposeDown.parse(["--cwd", project.base.path(percentEncoded: false)])
+        try await composeDown.run()
+
+        containers = try await ClientContainer.list()
+            .filter({
+                $0.configuration.id.contains(project.name)
+            })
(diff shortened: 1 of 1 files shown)


Sources/Container-Compose/Helper Functions.swift
```diff
@@ -21,37 +21,37 @@
...
-import Yams
 import Rainbow
-import ContainerCommands
+import Yams
 
 /// Loads environment variables from a .env file.
 /// - Parameter path: The full path to the .env file.
 /// - Returns: A dictionary of key-value pairs representing environment variables.
 public func loadEnvFile(path: String) -> [String: String] {
-    var envVars: [String: String] = [:]
```
Mcrich23: It looks like there was a lot of excess line changes due to a change in code indentation. Can you please fix this to reduce the number of lines being changed?

Mcrich23 (changes requested):
Thank you so much for fixing that. One more thing: can you please add a dynamic test for the full check bringing up the container?

Thank you!

PR #72, opened 2026-04-20
feat: Add compose build command

Luke-Parkin:
To align closer with docker/alternatives, add a `compose build` command which builds images but does not launch containers.

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeBuild.swift
+++ b/Sources/Container-Compose/Commands/ComposeBuild.swift
@@ -1,0 +1,166 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Morris Richman and the Container-Compose project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


Sources/Container-Compose/Commands/ComposeBuild.swift
```diff
@@ -0,0 +1,166 @@
...
+        let arch = String(((split ?? []).count >= 1 ? split?.last : nil) ?? "arm64")
+        commands.append(contentsOf: ["--os", os, "--arch", arch])
+
+        let cpuCount = Int64(service.deploy?.resources?.limits?.cpus ?? "2") ?? 2
+        let memoryLimit = service.deploy?.resources?.limits?.memory ?? "2048MB"
+        commands.append(contentsOf: ["--cpus", "\(cpuCount)", "--memory", memoryLimit])
+
+        print("\n----------------------------------------")
+        print("Building \(serviceName) -> \(imageTag)")
+        let buildCommand = try Application.BuildCommand.parse(commands)
```
Mcrich23: Can you please also have it pass down the logging flags to respect user decisions?

Mcrich23 (changes requested):
Hi @Luke-Parkin, thank you so much for patiently waiting for me to review this. Everything looks really good, but there's one thing regarding logging that needs to get fixed before I can merge it. Can you please take a look at my comment?

Sources/Container-Compose/Commands/ComposeBuild.swift
Mcrich23: The reason why we're on a fork instead of directly tracking main is to include a macro for this. Take a look at `ComposeUp` for more info.

Mcrich23 (approved):

Sources/Container-Compose/Commands/ComposeBuild.swift
Luke-Parkin: Hope that's what you're looking for! Thanks for the review.

PR #74, opened 2026-04-21
fix: Use var over let

Luke-Parkin:
Very sorry, looks like I broke the build with my last commit! (Still new to swift 😓)


Mcrich23:
Hey! No worries. It seems like I also got very excited and prematurely merged. This compilation error came from another PR that was just merged, so its not a big deal!

Mcrich23 (approved):
lgtm!

PR #75, opened 2026-04-22
fix(parsing): support map form of depends_on with condition syntax

24601:
## Problem

`container-compose` throws a `typeMismatch` error on any compose file that uses the map form of `depends_on`:

```yaml
depends_on:
  db:
    condition: service_healthy
```

This is valid Docker Compose spec (v2+) and is common in real-world stacks (Windmill, Keycloak, Traefik setups, etc). Only the list form worked:

```yaml
depends_on:
  - db
```

## Fix

In `Service.swift`, after failing to decode as `String` or `[String]`, attempt to decode as `[String: [String: String]]` and extract the map keys as the dependency list. Conditions are dropped since Apple Container has no equivalent orchestration-layer concept.

```swift
} else if let dependsOnMap = try? container.decodeIfPresent([String: [String: String]].self, forKey: .depends_on) {
    // Map form: depends_on: { db: { condition: service_healthy } }
    // Preserve dependency order; conditions are not applicable to Apple Container.
    depends_on = dependsOnMap.keys.sorted()
}
```

## Tests

Added `parseComposeWithDependenciesMapForm` covering:
- Two dependencies with different condition values (`service_healthy`, `service_started`)
- Both service names extracted correctly
- Count assertion

Both new and existing `depends_on` tests pass.

## Verified against

Windmill CE `docker-compose.yml` — previously failed at parse, now reaches image pull stage.

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Service.swift
+++ b/Sources/Container-Compose/Codable Structs/Service.swift
@@ -192,8 +192,14 @@ public struct Service: Codable, Hashable {
         
         if let dependsOnString = try? container.decodeIfPresent(String.self, forKey: .depends_on) {
             depends_on = [dependsOnString]
+        } else if let dependsOnArray = try? container.decodeIfPresent([String].self, forKey: .depends_on) {
+            depends_on = dependsOnArray
+        } else if let dependsOnMap = try? container.decodeIfPresent([String: [String: String]].self, forKey: .depends_on) {
+            // Map form: depends_on: { db: { condition: service_healthy } }
+            // Preserve dependency order; conditions are not applicable to Apple Container.
+            depends_on = dependsOnMap.keys.sorted()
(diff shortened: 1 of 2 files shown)


Cyb3rDudu:
Confirmed this fixes a real parser break on `main`. Minimal repro:

```yaml
services:
  db:
    image: alpine:latest
    command: ["sleep","60"]
  app:
    image: alpine:latest
    depends_on:
      db:
        condition: service_started
```

`main`: `Error: typeMismatch(Yams.Node.Sequence ... Expected to decode Sequence but found Node)` — full bail.
`pr-75`: parses fine, both containers start.

The map form is what `docker compose` config-converts everything to internally, so this one matters for any non-trivial migration. +1 from me.

Mcrich23 (approved):
looks good!

PR #78, opened 2026-05-03
fix(build): resolve dockerfile path relative to build context, not compose dir

Cyb3rDudu:
Per the Compose spec, `services.<svc>.build.dockerfile` is resolved relative to the resolved `build.context`, not to the directory of the compose file. Current `main` resolves both against `composeDirectory`, so any layout that puts the Dockerfile inside a subdirectory of the context (or uses the standard `context: ../foo` monorepo shape) fails:

```
$ container-compose build -f /tmp/ctx/compose-dir/test.yml
Error: dockerfile does not exist /tmp/ctx/compose-dir/Dockerfile
# (the Dockerfile is at /tmp/ctx/sub/Dockerfile, where context: ../sub points)
```

Changes:
- `ComposeBuild.swift`, `ComposeUp.swift`: resolve `dockerfile` against the resolved `contextURL` rather than `composeDirectory`. `context` itself is still resolved against `composeDirectory`, which was already correct.
- Adds three dynamic tests in `ComposeBuildTests.swift` covering: subdirectory context with default Dockerfile, parent-relative context (`../sub`), and custom dockerfile filename (`Dockerfile.dev`) inside the context.

All three tests fail on `main` with the "dockerfile does not exist" error and pass with this change. The 6 unrelated pre-existing test failures on `main` are unchanged.

Refs #34, supersedes #37.

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeBuild.swift
+++ b/Sources/Container-Compose/Commands/ComposeBuild.swift
@@ -131,14 +131,17 @@ public struct ComposeBuild: AsyncParsableCommand, @unchecked Sendable {
     ) async throws {
         let imageTag = service.image ?? "\(serviceName):latest"
 
-        var commands = [URL(fileURLWithPath: buildConfig.context, relativeTo: URL(fileURLWithPath: composeDirectory)).path]
+        // Per Compose spec: `context` is relative to the compose file's directory,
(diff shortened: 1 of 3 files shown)


Mcrich23 (approved):
lgtm!

Issue #76, opened 2026-05-03
container-compose always tries to pull from registry-1.docker.io, ignoring local images


Cyb3rDudu:
Pulling an image that already exists in the local store always reaches for Docker Hub.

```bash
container build -t myapp:local .   # ok
container run myapp:local           # ok
```

But:

```yaml
services:
  app:
    image: myapp:local
```

```
$ container-compose up -d
Pulling Image myapp:local...
Error: HTTP request to https://registry-1.docker.io/v2/myapp/manifests/local failed with response: 401 Unauthorized.
```

Workaround in use: stand up a local OCI registry on a free port (5000 is held by AirPlay on macOS), tag/`--scheme http` push, then reference the registry path. Heavy for what should just be local-build → local-run.

Could `container-compose` check the local image store first before reaching for the registry? `pull_policy: never|if_not_present|always` (the Compose v3+ standard) would also be ideal long-term.

container-compose 0.11.0, container 0.12.3, macOS 26.3.

Cyb3rDudu:
Resolved by #79, already merged into `main`. Thanks!

PR #79, opened 2026-05-03
fix(up): match locally-built multi-path images so pull is skipped

Cyb3rDudu:
`pullImage` only compared the last `/`-separated path component of each stored image reference against the requested image name. That works for registry-prefixed locals (`docker.io/library/nginx:latest` ↔ user typed `nginx:latest`), but misses two real cases:

- A locally-built image with a multi-path tag like `myorg/foo:local` matches itself only via exact reference, never via a path-component last-segment compare.
- The same flow but reversed: user types the registry-prefixed form and the local store has the short form.

Both fall into the registry-fetch path and hit `401 Unauthorized` from the default registry, even though the image is in the local store.

The match now also accepts:
- exact reference equality
- the stored reference being `<registry>/<imageName>` (registry-prefixed form of the requested name)

The existing last-segment match is preserved as a fallback so single-segment refs (`alpine:latest`) keep working.

Adds a dynamic test in `ComposeUpTests` that builds an image with a multi-path tag and then runs `up` referencing it. Fails on `main` with the 401, passes with this change. The 6 unrelated pre-existing failures on `main` are unchanged.

Refs #76.

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeUp.swift
+++ b/Sources/Container-Compose/Commands/ComposeUp.swift
@@ -594,7 +594,19 @@ public struct ComposeUp: AsyncParsableCommand, @unchecked Sendable {
 
     private func pullImage(_ imageName: String, platform: String?) async throws {
         let imageList = try await ClientImage.list()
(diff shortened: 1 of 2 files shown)


Mcrich23 (approved):
lgtm!

PR #82, opened 2026-05-04
fix(up): stop pinning a CPU core during foreground wait

### LINKED ISSUE #27: High CPU usage
I've noticed `container-compose up` is using 100% CPU (on a single core) after starting all the containers from `docker-compose.yml`. There's likely an issue with polling logs/state. A workaround is to exit the command after starting, the containers will keep running.

Cyb3rDudu:
`container-compose up` (foreground) holds 100% of one core after all containers have started. Reproducible with any compose file run without `-d`:

```yaml
services:
  idle:
    image: alpine:latest
    command: ["sleep", "60"]
```

```
$ container-compose up &
$ ps -o pcpu,comm -p $!
%CPU COMM
100.0 container-compose
```

The hot loop is `ComposeUp.swift::waitForever`:

```swift
for await _ in AsyncStream<Void>(unfolding: {}) {
    // This will never run
}
```

`AsyncStream<Void>(unfolding: () async -> Void?)` ends only when the closure returns `nil`. The empty closure returns `()`, which Swift auto-wraps as `.some(())` — never `nil` — so the stream emits an infinite sequence of `Void` values with no `await` between them. The for-loop iterates as fast as the runtime can deliver them. The `// This will never run` comment is inverted: the body runs constantly; only `fatalError("unreachable")` is unreachable.

Fix is `withUnsafeContinuation { _ in }`: the task suspends until the process is killed, with zero CPU. After this change the same repro shows `0.0 container-compose`.

Used `withUnsafeContinuation` rather than `withCheckedContinuation` because the latter logs a "continuation leaked" diagnostic at runtime — leaking is the intent here, since the contract is to wait until the process is killed.

Fixes #27.

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeUp.swift
+++ b/Sources/Container-Compose/Commands/ComposeUp.swift
@@ -194,9 +194,19 @@ public struct ComposeUp: AsyncParsableCommand, @unchecked Sendable {
     }
 
     func waitForever() async -> Never {
-        for await _ in AsyncStream<Void>(unfolding: {}) {
-            // This will never run
-        }
+        // `AsyncStream<Void>(unfolding: () async -> Void?)` ends only when the
+        // closure returns `nil`. An empty closure returns `()`, which Swift
+        // auto-wraps as `.some(())` — never `nil` — so the previous
+        // `for await _ in AsyncStream<Void>(unfolding: {})` produced an
+        // infinite stream of `Void` values with no `await` between them and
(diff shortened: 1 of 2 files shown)


Cyb3rDudu:
Quick bump — this is a one-line fix for the busy-wait CPU pinning in foreground mode. Happy to adjust if needed.

Mcrich23 (approved):
Thanks for the fix on this!

PR #83, opened 2026-05-04
fix(parsing): accept list-form environment (- KEY=value) alongside map form

### LINKED ISSUE #2: Add Environment Support
Including environment variables causes an error.

```
environment:
  - REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/data
```

> Error: typeMismatch(Yams.Node.Mapping, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "services", intValue: nil), _DictionaryCodingKey(stringValue: "registry", intValue: nil), CodingKeys(stringValue: "environment", intValue: nil)], debugDescription: "Expected to decode Mapping but found Node instead.", underlyingError: nil))

Cyb3rDudu:
Compose files using the list form of `environment:` fail to parse. Reporter's case from #2:

```yaml
services:
  registry:
    image: registry:2
    environment:
      - REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/data
      - REGISTRY_STORAGE_DELETE_ENABLED=true
```

The Compose spec allows both shapes for `services.<svc>.environment`:

```yaml
environment:                      environment:
  KEY: value                        - KEY=value
  OTHER: value     equivalent       - OTHER=value
                                    - INHERIT_FROM_HOST
```

`Service.swift` decodes only the map form (`[String: String]?`), so the list form throws `DecodingError.typeMismatch` and the whole compose file fails to load. List form is the older docker-classic style still used by Docker Registry, Sentry, GitLab, and many other official examples — comment thread on #2 has multiple users hitting it.

Fix tries the map form first (no behavior change for existing users), falls back to a small parser:
- `KEY=value` splits on first `=` (later `=` chars stay in the value, so DSN-style `postgres://u:p@h/db?sslmode=require` round-trips)
- Bare `KEY` reads from `ProcessInfo.processInfo.environment[KEY]`, falling back to empty string when unset (Compose's "inherit from host" shorthand)

Tests: 11 static cases covering both decode paths end-to-end (map form regression, list form, mixed forms across services, env key absent) plus targeted unit tests on the `parseEnvironmentList` helper (first-`=` split, DSN-style values, bare-key inheritance, duplicates, empty list).

Fixes #2. Fixes #65.

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Service.swift
+++ b/Sources/Container-Compose/Codable Structs/Service.swift
@@ -177,7 +177,21 @@ public struct Service: Codable, Hashable {
         restart = try container.decodeIfPresent(String.self, forKey: .restart)
         healthcheck = try container.decodeIfPresent(Healthcheck.self, forKey: .healthcheck)
         volumes = try container.decodeIfPresent([String].self, forKey: .volumes)
-        environment = try container.decodeIfPresent([String: String].self, forKey: .environment)
+
+        // `environment:` accepts both forms per the Compose spec:
+        //   environment:                          environment:
+        //     KEY: value                            - KEY=value
+        //     OTHER: value          equivalent       - OTHER=value
+        //                                            - INHERIT_FROM_HOST
+        // Try the map form first; fall back to list form.
+        if let asMap = try? container.decodeIfPresent([String: String].self, forKey: .environment) {
+            environment = asMap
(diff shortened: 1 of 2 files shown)


Cyb3rDudu:
Bumping this one too. I noticed #87 (environment parsing) was closed, so this should still be good to go. LMK if there's anything to tweak.

Mcrich23 (approved):
This looks great! Thank you.

PR #84, opened 2026-05-04
fix(up): normalize relative bind mount paths to absolute

### LINKED ISSUE #4: Support Relative Bind Mounts

Cyb3rDudu:
Compose files using relative bind mounts fail to start:

```yaml
services:
  app:
    image: node:latest
    volumes:
      - ./app:/app
```

```
app: Error: invalidArgument: "Invalid volume name './app': must match ^[A-Za-z0-9][A-Za-z0-9_.-]*$"
```

`configVolume` already computes an absolute `fullHostPath` for the directory-create path, but passes the original relative `source` to `container run -v`. The daemon then runs the value through its volume-name regex and rejects it.

Pulls the source-path resolution into a static helper `resolveBindMountSource(_:cwd:)` and wires `configVolume` through it. Already-absolute and `~`-prefixed paths pass through unchanged; relative paths get joined to `cwd` and standardized via `URL.standardizedFileURL`, collapsing `./` and `../` segments. The runtime side of the fix is the same as #42; this PR adds the unit tests Morris asked for there.

Tests: 9 static cases for the helper covering `./foo`, bare relative, `../foo`, intermediate `..`, absolute pass-through, `~/foo` literal preservation, `.//foo` double-slash, and the exact `./app` case from the issue.

Fixes #4. Supersedes #42 (credit to @ttys3 for the original approach).

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeUp.swift
+++ b/Sources/Container-Compose/Commands/ComposeUp.swift
@@ -685,6 +685,25 @@ public struct ComposeUp: AsyncParsableCommand, @unchecked Sendable {
         return imageToRun
     }
 
+    /// Normalize a bind-mount source path to an absolute, `.`/`..`-collapsed
+    /// host path suitable for passing to `container run -v`.
+    ///
+    ///   - `./foo`, `foo/bar`, `../foo`  → `<cwd>/foo`, `<cwd>/foo/bar`, `<parent>/foo`
+    ///   - `/abs/foo`                    → unchanged
+    ///   - `~/foo`                       → kept as-is (apple/container handles tilde
+    ///                                     expansion itself; FileManager would
+    ///                                     not, so leaving it literal preserves
+    ///                                     the existing pre-fix behavior)
+    static func resolveBindMountSource(_ source: String, cwd: String) -> String {
+        // Already-absolute or tilde-prefixed paths pass through; the daemon
+        // (or FileManager for the directory-create branch) handles them.
+        if source.starts(with: "/") || source.starts(with: "~") {
+            return source
+        }
+        let joined = "\(cwd)/\(source)"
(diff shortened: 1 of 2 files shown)


Cyb3rDudu:
Last one — just a nudge on the bind mount normalization. There's some adjacent work in #94 (file mounts) but this PR handles the relative path case separately. Let me know if you'd prefer these combined.

Mcrich23:
@Cyb3rDudu thank you for all of the PRs. I just merged #94 and it seems to have created a merge conflict. If you fix that, I would be happy to merge it in! Hoping to release 0.12.0 with all of these changes tomorrow.

Cyb3rDudu:
Superseded by #94 which resolves relative bind mount paths via `resolvedPath(for:relativeTo:)` in the refactored `composeVolumeToRunArgs()`. The core normalization is the same; #94 also added file mount forwarding on top.

PR #85, opened 2026-05-04
Expose core library target and expand Compose parser coverage

sf-jin-ku:
## Summary

- expose the core Compose implementation as a reusable SwiftPM product
- add backwards-compatible parsing for common Compose schema variants
- cover service/network/dependency forms without changing runtime behavior beyond parser compatibility

## Notes

This PR is intentionally static-first. Runtime orchestration changes are split into later PRs so parser compatibility can be reviewed independently.

## Verification

```bash
swift test --filter 'DockerComposeParsingTests|ServiceDependencyTests|NetworkConfigurationTests'\nswift test --filter Container_Compose_StaticTests\n```

PR #86, opened 2026-05-04
Expose SwiftPM products

sf-jin-ku:
## Summary
- expose ContainerComposeCore as a reusable SwiftPM library product
- expose the existing executable target as the container-compose product

## Notes
Split out of #85 so the package surface change can be reviewed separately from parser compatibility changes.

## Verification
- git diff --check
- swift build --product ContainerComposeCore
- swift build --product container-compose

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,6 +6,10 @@ import PackageDescription
 let package = Package(
     name: "Container-Compose",
     platforms: [.macOS(.v15)],
+    products: [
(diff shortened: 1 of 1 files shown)


sf-jin-ku:
Closing this split because the compatibility work does not need to expose ContainerComposeCore as a SwiftPM library product. Keeping the remaining parser/compatibility PRs scoped to Compose behavior only: #87, #88, #89, #90, and #91.

PR #97, opened 2026-05-29
feat(up): use Apple container DNS when domain registered, /etc/hosts fallback otherwise

### LINKED ISSUE #81: Use Apple container's built-in DNS for inter-container resolution when domain is registered
Today, services in the same compose project can't resolve each other by name — `getent hosts db` from inside `app` returns nothing, so configs that use `DB_HOST: db` (the docker-compose default) just don't work. Users have to pin IPs into env vars or manually script `extra_hosts`.

Apple `container` already ships a DNS server that the daemon's bridge gateway answers on, for any domain registered via `container system dns create <domain>`. So we can register a domain per project, run containers with dotted names like `db.<domain>`, and get real DNS for free:

- Container is named `<service>.<domain>` (Apple's convention from apple/container#800)
- `--dns-domain <domain>` is passed to `container run`, which writes `nameserver <gateway>` + `domain <domain>` into the container's `/etc/resolv.conf`
- Both `db` (short, via the implicit search list) and `db.<domain>` (FQDN) resolve through the daemon's DNS server

In `ComposeUp.run()` it's just: derive a DNS label from `projectName` (lowercased, `[a-z0-9-]`, ≤63 chars), shell out to `container system dns list` to check whether it's registered, and use the path above when it is. When it isn't, current behavior is unchanged plus a one-time printed note pointing at `sudo container system dns create <domain>`.

**Tested with apple/container 0.12.3 on macOS 26.** Two alpine containers `db.testdomain` and `app.testdomain` started with `--dns-domain testdomain`. From inside `app`:

```
/etc/resolv.conf:
    nameserver 192.168.65.1
    domain testdomain

getent hosts db.testdomain   → 192.168.65.62
getent hosts db              → 192.168.65.62
```

No `/etc/hosts` mutation, no per-container exec, survives container restarts (daemon updates the registration when IPs change).

**The catch:** `sudo container system dns create <domain>` is one-time per project and requires admin. For users who refuse sudo, behavior is unchanged.

I have a working implementation if you want it — ~70 net lines in `ComposeUp.swift`, a small sanitizer helper, ~15 unit tests, and one dynamic test that auto-skips if no domain is pre-registered. Happy to send a PR.

Cyb3rDudu:
Closes #81

When a project's sanitized name is already registered via `container system dns create`, containers are named `<svc>.<domain>` and get `--dns-domain` so the daemon's DNS server resolves peers. The /etc/hosts cross-patcher is skipped on this path. When the domain isn't registered, behavior is unchanged — legacy dashed names, /etc/hosts patching, and a one-time note with the sudo command.

ComposeDown tries all candidate name shapes (legacy dashed, dotted DNS, explicit `container_name`) so teardown works regardless of which mode created the containers.

Changes:
- `sanitizeDnsDomain` / `dnsListContainsDomain` — pure helpers for deriving a DNS label from the project name and checking registration
- DNS detection in `run()` — shells out to `container system dns list`, sets `dnsAvailable`
- `configService()` — passes `--name <svc>.<domain>` + `--dns-domain` when DNS is available
- `crossPatchHostsForService` — /etc/hosts fallback, gated on `!dnsAvailable`
- `stopOldStuffByName` + ComposeDown multi-shape cleanup

Tested with container 0.12.3 on macOS 26. Two alpine services with `name: dnstest` — dotted names resolve via getent, /etc/hosts untouched, resolv.conf carries the domain. Auto-skips when no domain is pre-registered.

Static tests: 11 tests covering sanitization rules and dns list parsing.
Dynamic test: end-to-end DNS path verification (skips gracefully without registered domain).

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeDown.swift
+++ b/Sources/Container-Compose/Commands/ComposeDown.swift
@@ -119,38 +119,41 @@ public struct ComposeDown: AsyncParsableCommand {
 
     private func stopOldStuff(_ services: [(serviceName: String, service: Service)], remove: Bool) async throws {
         guard let projectName else { return }
+        // The DNS path uses a dotted name `<svc>.<dnsDomain>`. Compute the
+        // domain the same way ComposeUp does so we can stop containers from
+        // either mode (and from a previous run that used a different mode).
+        let dnsDomain = ComposeUp.sanitizeDnsDomain(projectName)
 
         for (serviceName, service) in services {
-            // Respect explicit container_name, otherwise use default pattern
-            let containerName: String
-            if let explicitContainerName = service.container_name {
-                containerName = explicitContainerName
-            } else {
-                containerName = "\(projectName)-\(serviceName)"
+            var candidates: [String] = ["\(projectName)-\(serviceName)"]
+            if let dnsDomain { candidates.append("\(serviceName).\(dnsDomain)") }
+            if let explicit = service.container_name, !candidates.contains(explicit) {
+                candidates.append(explicit)
             }
 
-            print("Stopping container: \(containerName)")
-            
             let client = ContainerClient()
-            
-            guard let container = try? await client.get(id: containerName) else {
-                print("Warning: Container '\(containerName)' not found, skipping.")
-                continue
-            }
-
-            do {
-                try await client.stop(id: container.id)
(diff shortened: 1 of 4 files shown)


Mcrich23:
@Cyb3rDudu so sorry for the delay. Cleaning out the backlog right now, can you please resolve the conflicts with main?

PR #99, opened 2026-06-07
feat(commands): add --project-directory flag for alternate working directory

w1zdun:
Docker Compose ships --project-directory to override the working directory independently of the process CWD. Needed when running compose from a parent dir with -f pointing to a nested compose file.

Resolution priority: --project-directory > process.cwd > currentDirectoryPath. Applies to up, down, and build commands via shared ComposeFileOptions.


Mcrich23:
I like this addition. Can you rebase onto main and move the directory resolution into `ComposeProjectOptions`? Please add tests covering `--project-directory` with `-f` and relative paths. #109 also changes how the host working directory is selected. Can you compare the two so both options use the same directory resolution instead of introducing separate implementations?

PR #101, opened 2026-06-07
fix(run): do not pass unsupported --hostname to container run

w1zdun:
The container CLI has no --hostname flag, so any service declaring 'hostname:' made the whole run command fail with "Unknown option '--hostname'". The field is now ignored with an explanatory note, matching how the unsupported restart policy is handled.


Cyb3rDudu:
closing — superseded by #119 (hostname is handled now, warns and skips it).

PR #102, opened 2026-06-07
Fix/variable interpolation

w1zdun:
Compose-file values support `${VAR}` / `${VAR:-default}` interpolation, but three consumers used the raw string, so common docker-compose patterns reached the runtime literally:

  - `build.context: ${REPOS_PATH:-..}/service` → "context dir does not exist `.../${REPOS_PATH:-..}/service`". New `resolveBuildPaths` helper interpolates context and dockerfile
  values stay literal, matching docker compose) and keeps the service-overrides-env-file precedence.


Mcrich23:
These fixes look useful. Can you rebase onto main and put container-name resolution in the shared project code so `up` and `down` use the same names, including DNS names? There's overlap with #153 in how service environment values are resolved. Can you compare the two implementations so we preserve the overrides this PR fixes when combining them? One clarification: this doesn't appear to address the command interpolation reported in #140.

PR #103, opened 2026-06-07
fix(parsing): shell-lex string-form command and entrypoint into argv

w1zdun:
Per Compose semantics, a string command like
'sh -c "npm install && npm run build:watch"' must be tokenized shell-style into ["sh", "-c", "npm install && npm run build:watch"]. It was previously wrapped as a single-element array, so the whole string was passed as one argument to the image entrypoint, e.g. node failing with "Cannot find module '/app/.../sh -c npm install...'". Adds a small POSIX-like lexer (whitespace, quotes, backslash escapes; no expansion) used for both command and entrypoint string forms.


Mcrich23:
Thank you for fixing this. The lexer currently accepts unmatched quotes and drops a trailing backslash. Can you add tests for those cases and correct the handling? I'd also like a dynamic test confirming a quoted command actually executes with the expected arguments.

PR #104, opened 2026-06-07
feat(commands): add -p/--project-name flag and docker compose projec name resolution

w1zdun:
Resolve the project name with the same precedence as docker compose: -p/--project-name flag > COMPOSE_PROJECT_NAME (process env > .env) > top-level 'name' field (now with ${VAR:-default} interpolation) > directory name.

Previously a name like ${COMPOSE_PROJECT_NAME:-myproj} was used literally, producing invalid container names. ComposeDown also gains .env file loading so 'down' resolves the same names as 'up'.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ DerivedData/
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
 .netrc
 .vscode/launch.json
+/tmp/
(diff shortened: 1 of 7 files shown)


Mcrich23:
I like this change, but please merge in main since there was a major version upgrade.

Mcrich23:
I still like this change. Main now has shared project resolution, so can you move this there and resolve the conflicts? Please test that starting and stopping use the same project name, including when `-f` points outside the current directory.

PR #105, opened 2026-06-07
feat(services): deterministic service selection for partial up/down

w1zdun:
Service selection for 'up <service>' was nondeterministic: topoSort recorded dependedBy on a value copy that only survived the first visit, and the input order came from a Dictionary, so 'up app' sometimes started its dependencies and sometimes did not. The filter also only checked direct dependents, losing transitive dependencies.

- configuredServices(from:) — deterministic (alphabetical) service order from the decoded YAML mapping
- expandServiceSelection(requested:services:) — transitive depends_on closure, matching 'docker compose up <service>' semantics
- down <service> now stops only the named services (docker compose parity, no dependency expansion)
- removed the broken Service.dependedBy field (dead after the filter rewrite)

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Service.swift
+++ b/Sources/Container-Compose/Codable Structs/Service.swift
@@ -97,10 +97,7 @@ public struct Service: Codable, Hashable {
 
     /// Allocate a pseudo-TTY (-t flag for `container run`)
     public let tty: Bool?
-    
-    /// Other services that depend on this service
-    public var dependedBy: [String] = []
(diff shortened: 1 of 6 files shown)


Mcrich23:
Again, please share issues first documenting these problems. It is incredibly difficult for me to review a PR you make where you assert an issue with no way for me to verify that you are correct.

Cyb3rDudu:
closing — superseded by #119 + #126 — service.selectServices covers transitive selection.

PR #106, opened 2026-06-07
Feat/native volumes

w1zdun:
Named volumes were emulated as host directories under `~/.containers/Volumes/...` bind-mounted via virtiofs. Containers whose entrypoint `chown`s its data directory (MySQL,
  Postgres, most databases) die on virtiofs (`chown: Operation not permitted`). Apple container ≥ 0.12 supports volumes natively, exposed by the pinned dependency via
  `ClientVolume`.

  Changes:

  - Top-level `volumes:` are created as native volumes named `<project>_<key>` (docker compose naming parity); `external: true` volumes are never created; explicit `name:` is
  honored. Creation is idempotent — existing volumes are detected via `ClientVolume.list()` and an "already exists" error from create is tolerated.
  - Named volumes referenced by services but not declared top-level are created on demand (docker compose errors here; this tool has historically been tolerant, so that behavior
  is kept).
  - `composeVolumeToRunArgs` passes `-v <nativeName>:<destination>[:mode]` for named volumes — this also fixes a pre-existing bug where the named-volume branch mounted the
  *parent* of the destination (`deletingLastPathComponent`). Bind-mount behavior is byte-for-byte unchanged, covered by the existing tests.
  - New `down -v`/`--volumes` flag removes the project's named volumes (including on-demand ones, skipping external). Containers are removed first, since a volume cannot be
  deleted while a container references it — matching `docker compose down`, which always removes containers.

  Note: unlike docker, Apple container does not populate a freshly created volume from the image's directory content — that is a runtime limitation, not addressed here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ DerivedData/
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
 .netrc
 .vscode/launch.json
+/tmp/
--- a/Sources/Container-Compose/Commands/ComposeDown.swift
+++ b/Sources/Container-Compose/Commands/ComposeDown.swift
@@ -38,6 +38,9 @@ public struct ComposeDown: AsyncParsableCommand {
     @Argument(help: "Specify the services to stop")
     var services: [String] = []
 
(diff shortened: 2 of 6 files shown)


Cyb3rDudu:
closing — superseded by #119 (native named volumes landed via composeNamedVolumeName).

PR #107, opened 2026-06-07
fix(tests): widen waitForever CPU test window to eliminate parallel-suite flakiness

w1zdun:
The regression test for #27 measures process-wide user CPU (getrusage(RUSAGE_SELF)) over a 200ms window with a 50,000 µs threshold. swift-testing runs all other suites in parallel in the same process, and their overhead alone was observed at ~70,000 µs — tripping the threshold with no busy-loop present.

The bug being guarded against pins a full core (~1 µs CPU per µs of wall-clock), so the threshold must scale with the window while suite noise stays roughly constant (the rest of the static suite finishes within the first ~100ms). Widen the window to 500ms and set the threshold to half a core (250,000 µs): ~3x above observed noise, ~2x below a pinned core.

Verified: 3 consecutive full static-suite runs pass; with the original busy-loop temporarily reintroduced the test fails (528,314 µs observed).

### DIFF
--- a/Tests/Container-Compose-StaticTests/WaitForeverCpuTests.swift
+++ b/Tests/Container-Compose-StaticTests/WaitForeverCpuTests.swift
@@ -32,18 +32,22 @@ struct WaitForeverCpuTests {
     /// Regression for #27: `waitForever()` must suspend, not busy-loop.
     ///
     /// Method: take a `getrusage(RUSAGE_SELF)` snapshot, spawn a child Task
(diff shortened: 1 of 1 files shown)


Mcrich23:
This issue has not been observed in CI or in my testing. Can you please document reproducible steps?

Cyb3rDudu:
closing — superseded by #126, which fixed the cpu test with an idle-baseline approach.

PR #109, opened 2026-06-16
feat(compose): add --env-file and --workdir flags

771991673:
Add --env-file and --workdir/-w as top-level options on ComposeFileOptions, matching Docker Compose CLI semantics. These flags were previously inherited from Flags.Process (the container runtime dependency) which conflated compose-level concerns with container runtime flags.

- Add --env-file to ComposeFileOptions for specifying alternate .env files
- Add --workdir/-w to ComposeFileOptions for alternate working directory
- Remove Flags.Process from ComposeUp, ComposeBuild, ComposeDown
- ComposeDown no longer needs import ContainerCommands
- Add 7 parsing tests covering all subcommands and defaults

### DIFF
--- a/Sources/Container-Compose/Commands/ComposeBuild.swift
+++ b/Sources/Container-Compose/Commands/ComposeBuild.swift
@@ -46,12 +46,9 @@ public struct ComposeBuild: AsyncParsableCommand, @unchecked Sendable {
     var noCache: Bool = false
 
     @OptionGroup
-    var process: Flags.Process
-
-    @OptionGroup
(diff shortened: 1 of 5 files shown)


Mcrich23:
Separating project options from the container runtime flags makes sense. Can you update this to use `ComposeProjectOptions`? #99 adds `--project-directory`, while this PR adds `--workdir` for the same host directory choice. Can you compare them and have both use the same resolution, with tests showing which Compose and environment files are loaded when running from another directory?

Issue #118, opened 2026-06-25
Track Compose compatibility gaps found with apple/container 1.0.0


thromel:
## Summary

This tracks a small set of Compose compatibility gaps found while testing `container-compose` 1.0.0 with `apple/container` 1.0.0 on macOS 26.5.1.

The goal is to keep the work contribution-friendly: each linked issue is narrow enough to fix or validate independently.

This tracker is for Compose compatibility work in `container-compose`. Runtime-level fixes should happen upstream in `apple/container`; the main one from this test run is container name/service DNS.

## Environment used for the repros

```markdown
- OS: macOS 26.5.1 (25F80)
- Xcode: 26.5 (17F42)
- container: 1.0.0 (release, commit ee848e3)
- container-compose: 1.0.0 installed via Homebrew
- Architecture: arm64
```

## Issues

- [ ] #114
  Services on the same Compose network cannot resolve service names.

- [x] #115
  One-shot services are reported as startup timeouts and failures can exit 0.

- [x] #116
  Parse Compose service network object form for aliases.

- [x] #117
  `healthcheck` and `depends_on: condition: service_healthy` are silently ignored.

- [x] #52
  Named volume workaround fails; native `container run -v data:/data ...` now appears to support named volumes directly.

## Upstream dependencies / related Apple Container issues

- apple/container#856
  Container name DNS resolution currently fails in a plain `container` repro on `container` 1.0.0. This likely affects #114 and is the best first candidate for an upstream `apple/container` contribution.

- apple/container#1752
  Tracks log retrieval semantics needed by Compose-compatible orchestrators. Not part of the first repro batch, but relevant for future `container-compose logs` work.

## Suggested contribution order

1. #115: fix one-shot service wait semantics and CLI exit status.
   This should be mostly internal `container-compose` behavior and is easy to regression-test.

2. #116: accept Compose network object syntax.
   Even if aliases cannot fully work until service discovery improves, the YAML shape should decode cleanly.

3. #52: switch named-volume handling to native `container` named volumes where supported.
   The current workaround looks stale against `container` 1.0.0.

4. #117: either implement healthcheck/dependency behavior or emit a clear unsupported-feature error.
   Silent ignore is the risky part.

5. #114: track Apple Container DNS/service discovery behavior.
   If upstream runtime support is required, the immediate improvement may be documentation, warning output, and tests that mark the limitation explicitly.

## Definition of done

- Minimal repro compose files for each issue are added as tests or fixtures.
[...]

Issue #117, opened 2026-06-25
healthcheck and depends_on service_healthy are silently ignored


thromel:
## Summary

`healthcheck` and `depends_on.condition: service_healthy` are parsed without an error, but the dependent service starts immediately and the healthcheck does not appear to be represented in the created container configuration.

## Environment

```markdown
- OS: macOS 26.5.1 (25F80)
- Xcode: 26.5 (17F42)
- container: 1.0.0 (release, commit ee848e3)
- container-compose: 1.0.0 installed via Homebrew
- Architecture: arm64
```

## Reproduction

`compose.yml`:

```yaml
services:
  web:
    image: nginx:alpine
    healthcheck:
      test: ["CMD", "wget", "-qO-", "http://127.0.0.1"]
      interval: 1s
      timeout: 1s
      retries: 5

  probe:
    image: busybox:latest
    depends_on:
      web:
        condition: service_healthy
    command: ["sh", "-c", "echo PROBE_STARTED"]
```

Run:

```sh
container-compose --file compose.yml up -d
container logs health-probe
container inspect health-web
```

## Actual behavior

`probe` starts immediately:

```text
PROBE_STARTED
```

`container inspect health-web` did not show healthcheck metadata or a health status field.

## Expected behavior

For Compose compatibility, one of these should happen:

1. `container-compose` implements healthcheck execution and waits for `service_healthy` before starting dependents.
2. If this requires upstream `apple/container` healthcheck primitives, `container-compose` reports a clear unsupported-feature warning/error instead of silently ignoring the condition.

Silent ignore is risky because stacks can appear to start while dependencies are not actually ready.

ryan106:
This also looks fixed (likely by #119, same as #115/#52) — `depends_on: condition:` is fully implemented in `ComposeUp.swift`'s `waitForDependencyConditions`, not silently ignored:

```swift
case ServiceDependency.serviceStarted:
    guard serviceStartStates[dependencyName] != nil else {
        throw ComposeError.dependencyNotStarted(serviceName, dependencyName)
    }
case ServiceDependency.serviceHealthy:
    guard serviceHealth[dependencyName] == true else {
        throw ComposeError.dependencyNotHealthy(serviceName, dependencyName)
    }
case ServiceDependency.serviceCompletedSuccessfully:
    guard serviceStartStates[dependencyName] == .completed else {
        throw ComposeError.dependencyNotCompleted(serviceName, dependencyName)
    }
default:
[...]

Issue #116, opened 2026-06-25
Parse Compose service network object form for aliases


thromel:
## Summary

`container-compose` fails to decode the Compose object form for service networks, which is required for aliases and per-network service options.

## Environment

```markdown
- OS: macOS 26.5.1 (25F80)
- Xcode: 26.5 (17F42)
- container: 1.0.0 (release, commit ee848e3)
- container-compose: 1.0.0 installed via Homebrew
- Architecture: arm64
```

## Reproduction

`compose.yml`:

```yaml
services:
  web:
    image: nginx:alpine
    networks:
      appnet:
        aliases:
          - web-alias

  probe:
    image: busybox:latest
    depends_on:
      - web
    networks:
      - appnet
    command: ["sh", "-c", "wget -qO- http://web-alias"]

networks:
  appnet:
```

Run:

```sh
container-compose --file compose.yml up -d
```

## Actual behavior

```text
Error: DecodingError.typeMismatch: expected value of type Sequence. Path: services.web.networks. Debug description: Expected to decode Sequence but found Node instead.
```

## Expected behavior

Compose supports both forms:

```yaml
networks:
  - appnet
```

and:

```yaml
networks:
  appnet:
    aliases:
      - web-alias
```

`container-compose` should parse the object form and either:

- apply supported per-network service settings, or
- produce a targeted unsupported-field warning while continuing for settings it can support.

Aliases may depend on the broader service-discovery/DNS work, but the YAML decoder should accept the Compose object shape.

thromel:
Closing as completed by #119. Current main parses the Compose service network object form, including aliases; NetworkConfigurationTests covers the YAML shape from this issue.

Issue #115, opened 2026-06-25
One-shot services are reported as startup timeouts and failures can exit 0


thromel:
## Summary

Short-lived services that exit successfully are reported as startup timeouts, and some failed starts still cause `container-compose up` to exit with status `0`.

This makes one-shot Compose jobs and CI usage hard to trust.

## Environment

```markdown
- OS: macOS 26.5.1 (25F80)
- Xcode: 26.5 (17F42)
- container: 1.0.0 (release, commit ee848e3)
- container-compose: 1.0.0 installed via Homebrew
- Architecture: arm64
```

## Reproduction: successful one-shot service

`compose.yml`:

```yaml
services:
  cmdonly:
    image: busybox:latest
    command: ["sh", "-c", "echo CMD_OK"]
```

Run:

```sh
container-compose --file compose.yml up -d cmdonly
container list --all | grep cmdonly
container logs feature-isolation-cmdonly
```

## Actual behavior

`container-compose` reports:

```text
Error Domain=ContainerWait Code=1 "Timed out waiting for container 'feature-isolation-cmdonly' to be running."
```

But the container actually ran successfully and exited:

```text
feature-isolation-cmdonly  docker.io/library/busybox:latest  linux  arm64  stopped
```

Logs:

```text
CMD_OK
```

The same happened for a locally built one-shot image:

```text
build-built  local-compose-build-test:latest  linux  arm64  stopped
```

Logs:

```text
from-build-arg
```

## Expected behavior

If a service process exits before reaching a long-running `running` state, `container-compose` should inspect the exit result:

- exit code `0`: treat as a successful completed one-shot service
- non-zero exit code: report failure and return a non-zero CLI exit status
- still booting/running: continue the existing wait behavior

This would match common Compose usage for migration jobs, probes, build checks, and test containers.

## Related status-code issue

In a separate named-volume start failure, `container-compose up -d` printed an internal error but still returned shell exit code `0`. That should return non-zero so scripts and CI can detect failure.

ryan106:
This looks fixed by #119 (merged yesterday) — I tested both halves of the repro against current `main`:

**Successful one-shot:**
```yaml
services:
  cmdonly:
    image: busybox:latest
    command: ["sh", "-c", "echo CMD_OK"]
```
`container-compose --file compose.yml up -d cmdonly` completes cleanly, no "Timed out waiting..." error, container reaches `.completed` state correctly.

**Failing one-shot:**
```yaml
services:
  failer:
[...]

PR #119, opened 2026-06-25
Improve Apple container compose compatibility

### LINKED ISSUE #1839: [Request]: Add network-scoped aliases for container network attachments
### Feature or enhancement request details

Please add network-scoped aliases for container network attachments.

#### Use case

Higher-level tools such as Compose adapters need to attach additional DNS names to a container on a specific network without changing the container's management name.

For example, a Compose service may have:

```yaml
services:
  postgres:
    networks:
      backend:
        aliases:
          - db
          - database
```

The runtime needs a way to represent those names as aliases for the same network attachment so peer containers on `backend` can resolve `postgres`, `db`, and `database` to the same attachment IP. The aliases should be scoped to the network attachment, not global host DNS.

This complements the broader container-to-container DNS discovery request in #1809.

#### UX / CLI shape

Use the existing comma-separated `--network` property syntax and allow repeated `alias=NAME` entries:

```sh
container run --network backend,alias=db,alias=database ...
container create --network backend,alias=db ...
```

Each `alias=NAME` applies only to that network attachment. Multiple networks can therefore carry different aliases for the same container.

#### Proposed design

At a high level:

- Extend network attachment options with an aliases array, alongside the existing primary hostname, MAC address, and MTU options.
- Parse repeated `alias=NAME` values from the existing `--network` option grammar.
- Carry aliases through the container API client, persisted attachment configuration, network helper allocation request, and returned attachment metadata.
- Treat the primary hostname and aliases as names for the same allocation in the hostname database.
- Validate duplicate names across active allocations before creating the container, so two attachments cannot claim the same hostname or alias in the same network.
- Normalize aliases using the same hostname lookup behavior as primary attachment hostnames, including case and trailing-dot handling, rather than introducing a stricter validator in this feature.
- Release all names for an allocation when the network attachment is released.

The draft implementation in #1815 follows this shape:

```sh
container run --network backend,alias=db,alias=database ...
```

In live validation against a debug build, peer containers on the same network could resolve the primary container name and both aliases through the container-facing network DNS listener, and a duplicate `alias=db` failed with the expected duplicate-hostname error.

#### Code of Conduct

I agree to follow this project's Code of Conduct.

thromel:
## Summary

This improves compatibility with Apple `container` for several Compose workflows:

- decode `depends_on` map form and preserve dependency conditions
- wait for `service_started`, `service_healthy`, and `service_completed_successfully`
- parse service `networks` object form, including aliases
- emit Apple network alias properties when the linked Apple `container` command parser supports them, while preserving a warning/skip fallback for Apple Container 1.0.0
- warn and skip Compose `hostname:` because Apple `container run` 1.0.0 does not currently expose a hostname flag; proposed upstream support remains under discussion in apple/container#1811
- execute supported healthchecks (`CMD` and `CMD-SHELL`) with interval/start-period/retry handling
- create and mount named volumes through native `container volume` support instead of rewriting them into host paths
- surface failed detached `container run` commands as errors, which makes one-shot failures fail `compose up`

Related issues: #52, #115, #116, #117, and tracker #118.

Apple-side service discovery work is tracked upstream in apple/container#1809. Current upstream PR chain:

- apple/container#1810: normalized/trailing-dot hostname lookup, signed and ready for review
- apple/container#1811: optional native `container run --hostname`, signed and under maintainer discussion; aliases may cover most Compose cases
- apple/container#1813: DNS forwarding/message groundwork only; container-facing listener startup was removed after vmnet gateway bind validation failed
- apple/container#1815: network attachment aliases via `--network name,alias=...`, signed, closes apple/container#1839

## Testing

Static/build checks:

```sh
git diff --check
swift build
swift test --filter EntrypointCommandTests
swift test --filter Container_Compose_StaticTests --skip WaitForeverCpuTests
```

The static test command passed 141 tests. `WaitForeverCpuTests` was skipped because the existing CPU-threshold test is noisy on this machine and unrelated to these changes.

Manual Apple `container` smoke tests using the locally built `Container-Compose` binary:

- one-shot success dependency completes and unblocks dependent startup
- one-shot failure exits nonzero and fails `up -d`
- native named volume persists data and can be read from a separate `container run`
- service network object form attaches to the requested network and emits an alias fallback warning against Apple Container 1.0.0
- passing healthcheck unblocks dependent service startup
- failing healthcheck prevents dependent service startup
- service with `hostname:` emits a warning, does not pass unsupported `--hostname`, and completes successfully (`HOSTNAME_OK` logged)

All dynamic test containers, networks, and volumes were cleaned up after the runs.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "88a8c9d7a638866eedf9a6130a6e13bdb1a21e134677e63d45429b307709c569",
+  "originHash" : "79acd8672b13f4d188e7a4c5d86093adfa71839a04733296d715993c9f62ce99",
   "pins" : [
     {
       "identity" : "async-http-client",
--- a/Package.swift
+++ b/Package.swift
@@ -25,6 +25,10 @@ let package = Package(
                     package: "container"
                 ),
                 .product(
+                    name: "ContainerXPC",
+                    package: "container"
+                ),
+                .product(
                     name: "ArgumentParser",
                     package: "swift-argument-parser"
                 ),
--- a/Sources/Container-Compose/Codable Structs/Healthcheck.swift
+++ b/Sources/Container-Compose/Codable Structs/Healthcheck.swift
@@ -25,6 +25,17 @@ import Foundation
 
 /// Healthcheck configuration for a service.
 public struct Healthcheck: Codable, Hashable {
+    private static let durationUnits: [String: TimeInterval] = [
+        "ns": 0.000000001,
+        "us": 0.000001,
+        "µs": 0.000001,
+        "ms": 0.001,
+        "s": 1,
+        "m": 60,
+        "h": 3600,
+    ]
+    private static let durationRegex = try! NSRegularExpression(pattern: #"([0-9]+(?:\.[0-9]+)?)(ns|us|µs|ms|s|m|h)"#)
+
     /// Command to run to check health
     public let test: [String]?
     /// Grace period for the container to start
@@ -98,19 +109,8 @@ public struct Healthcheck: Codable, Hashable {
             return seconds
         }
 
-        let units: [String: TimeInterval] = [
-            "ns": 0.000000001,
-            "us": 0.000001,
-            "µs": 0.000001,
-            "ms": 0.001,
-            "s": 1,
-            "m": 60,
-            "h": 3600,
-        ]
-        let pattern = #"([0-9]+(?:\.[0-9]+)?)(ns|us|µs|ms|s|m|h)"#
-        let regex = try! NSRegularExpression(pattern: pattern)
         let range = NSRange(value.startIndex..<value.endIndex, in: value)
-        let matches = regex.matches(in: value, range: range)
+        let matches = Self.durationRegex.matches(in: value, range: range)
         guard !matches.isEmpty else {
             return defaultValue
         }
@@ -123,7 +123,7 @@ public struct Healthcheck: Codable, Hashable {
             else {
                 return total
             }
-            return total + amount * (units[String(value[unitRange])] ?? 0)
+            return total + amount * (Self.durationUnits[String(value[unitRange])] ?? 0)
         }
     }
 }
--- a/Sources/Container-Compose/Codable Structs/Healthcheck.swift
+++ b/Sources/Container-Compose/Codable Structs/Healthcheck.swift
@@ -21,6 +21,7 @@
 //  Created by Morris Richman on 6/17/25.
 //
 
+import Foundation
 
 /// Healthcheck configuration for a service.
 public struct Healthcheck: Codable, Hashable {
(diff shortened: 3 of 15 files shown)


Cyb3rDudu (changes requested):
Great PR - this meaningfully closes the gap with Apple container, and the parsing/arg-building test coverage is solid. I'd like to merge this, but I think two items should be addressed first since they're regressions on existing workflows rather than missing features. The rest can be follow-ups.

Blockers (before merge)

1. up <service> now hard-fails on transitive dependencies
Sources/Container-Compose/Commands/ComposeUp.swift:160

The service filter only keeps a service if it's requested or a direct dependent (via dependedBy, which topoSortConfiguredServices populates one level deep). For a chain c → b → a, compose up c keeps c and b but drops a. The new waitForDependencyConditions then throws dependencyNotStarted("b", "a"). Before this PR the missing grandparent was tolerated, so this is a regression on a core workflow.

Fix: expand the filter to keep the full transitive dependency closure of the requested services (walk depends_on recursively), or have service_started start a missing dependency on demand instead of throwing.

2. up -d silently reports success when a long-running service crashes
Sources/Container-Compose/Commands/ComposeUp.swift:336

container run -d returns exit 0 on backgrounding, so the guard at :726 passes. If the container then crashes within the poll window, it reaches .stopped and waitUntilServiceStarted records .completed — the container's real exit code is never inspected. This undermines the PR's own goal of surfacing failed detached runs. A db that exits non-zero ~1s after start will print success and launch dependents against a dead service.

Fix: for a .stopped long-running service, inspect the container's actual exit status rather than assuming success.

Follow-ups (file as issues, don't block)

- retries is total attempts, not consecutive failures (ComposeUp.swift:797) — a healthy-but-slow service (e.g. a DB needing ~45s) is declared unhealthy. Compose treats retries as consecutive failures.
- healthcheck.timeout is parsed but never applied (ComposeUp.swift:806) — a hanging probe blocks streamCommand indefinitely instead of being bounded.
- service_completed_successfully race in attached mode (ComposeUp.swift:730) — the detached Task + immediate poll can sample a one-shot as .running before it exits, breaking its dependents. Timing-dependent; worth a known-limitation note.

Minor (cleanup)

- supportsNetworkAliases() is re-evaluated on every networkRunArg call (:295) — cache it in a static let.
- Healthcheck.parseDuration recompiles the regex on every call (Healthcheck.swift:58) — hoist to static let.
- The detach/attached branches duplicate handleOutput, the start banner, and the streamCommand call (:716) — extract the shared part and branch only on error handling.

Happy to help with any of these. The core work here is solid - just want to land it without the two regressions.

Cyb3rDudu:
Separate from the items in the review above - wanted to flag one thing on the volume handling specifically.

The native-volume switch is the right call, and I like that it respects `name` / `external` / `driver` / `labels` and replaces the old `createVolumeHardLink` host-dir dance. One gap though: the volume ends up named just `data` (or the explicit `name`), with no project prefix. Docker Compose names these `<project>_<source>` precisely so two stacks on the same machine don't collide on a shared name - as it stands, two projects both declaring `volumes: { data: ... }` would write into the same `data` volume.

Coincidentally #122 just opened and hits this same branch from the other direction: it uses `<project>_<source>` namespacing (which solves the collision) but drops the `name` / `external` / `driver` handling and leaves `createVolumeHardLink` running as dead code. I think your approach here is the more complete one — would you be open to borrowing #122's namespacing as the default name? Roughly:

```swift
let actualVolumeName = volumeConfig.name
    ?? volumeConfig.external?.name        // external: keep as-is, skip create
    ?? "\(projectName)_\(source)"         // fall back to project-namespaced
```
That keeps everything you've built and adds the isolation. Happy to help wire it up.

thromel:
Thanks for the detailed review. I pushed `bdc0a64` with the merge blockers fixed: `up <service>` now keeps the full dependency closure, stopped detached services check the actual init-process exit code, and default native volume names are project-scoped while preserving `name` / `external`. I also cached alias-support probing and hoisted the healthcheck duration regex. Static affected tests and `swift build` passed locally; I left the retry/timeout and attached one-shot race items as follow-ups per your note.

Cyb3rDudu:
@thromel I pulled `main` into this branch to clear the conflicts that landed after #97 merged (DNS, compose labels, the `ActivityClock` idle-wait, and the `stopOldStuff` → `stopExistingContainers` rename). The PR is `MERGEABLE` again.

The conflict was all in `ComposeUp.swift` - I merged both sides rather than picking one:
- the readiness wait keeps your `waitUntilServiceStarted` / `ServiceStartState` / `.stopped` exit-code check, and adopts main's `ActivityClock` idle + `maxWait` signature and `containerName(for:)`;
- the run path keeps your detach (exit-code guard) vs attached split, with an `ActivityClock` threaded through both branches so the idle wait sees pull progress;
- teardown uses main's `stopExistingContainers` 

Builds clean; 104 static tests pass.

Two notes:
1. I left out main's `runTask` teardown-on-wait-failure - it assumes a single wrapping `runTask`, which your detach/attached split doesn't have. Worth re-adding once the run path is unified. No regression vs. your branch; it just doesn't gain the cleanup.
2. Separately, my earlier note on `waitForInitExitCode` still stands - worth double-checking the `processIdentifier` (currently set to the container *name*, not a PID) and that `.exitCode` is always present, so a missing value can't default to `0` (= success).

Happy to merge once you've confirmed the XPC behavior in (2) — either a quick
smoke test (exit-0 vs non-zero) or a small fix. (1) is fine as a follow-up;
no need to block the merge on it.

Everything else looks good. One tiny optional tweak: the teardown bullet currently reads "teardown uses main's stopExistingContainers" with no why — you could append "(your stopOldStuff had no callers left after the merge)" so it's clear it wasn't an arbitrary choice, but that's cosmetic.

thromel:
Confirmed the XPC concern was valid: waiting only after the container reached `.stopped` was too late because the runtime client may already be gone. I pushed `5dfb26e`, which registers the `containerWait` task immediately after detached `container run` succeeds, then consumes that task if startup polling observes `.stopped`.

Smoke-tested locally with `busybox:latest`:
- `exit 0` one-shot: `up -d` succeeds.
- `exit 17` one-shot: `up -d` fails with `Service 'fail' failed to start ... status 17`.

Also reran `swift build` and `swift test --filter 'ServiceDependencyTests|ComposeVolumeTests|ActivityClockTests'`. The `runTask` teardown item still looks fine as a follow-up per your note.

Cyb3rDudu (approved):

PR #122, opened 2026-06-27
Map named volumes to native container volumes (fixes #96)

### LINKED ISSUE #96: volumes mounts parent directory than we expect, may tries to mount root directory
we have compose file like the following
```
services:
  redis:
    restart: always
    image: redis:7-alpine
    entrypoint: /usr/bin/docker-entrypoint.sh
    volumes:
      - redis-data:/data
#    healthcheck:
#      test: "redis-cli ping"
#      interval: 5s
#      retries: 20
    ports:
      - "6379:6379"
```

The container-compose tries to mount `~/.containers/Volumes/misskey/redis-data:/` results no entry point error.
Container compose should mount `~/.containers/Volumes/misskey/redis-data` to `/data` instead

https://github.com/Mcrich23/Container-Compose/blob/8c5b8161f9e912740d1a0225db4bdd9622a704d0/Sources/Container-Compose/Commands/ComposeUp.swift#L738-L739

I think this line causes the problem but I could not find why `deletingLastPathComponent` is added there

kylebrowning:
## Problem

The named-volume branch of `composeVolumeToRunArgs` bind-mounted a host directory and mounted it at the **parent** of the declared destination via `destination.deletingLastPathComponent()`. Two failure modes fall out of this:

1. **Mounts over `/` (issue #96).** A single-segment destination collapses to `/`. So `redisdata:/data` mounts the volume over the container's entire root filesystem, shadowing the image's entrypoint. Redis fails to start with:
   ```
   failed to find target executable docker-entrypoint.sh
   ```
2. **No chown on virtiofs.** Even for nested destinations, the host bind mount uses virtiofs, which the guest cannot `chown`. Images whose entrypoint chowns its data directory (postgres, redis, …) die with:
   ```
   chown: .: Operation not permitted
   ```

A side effect of (1) for nested paths (e.g. `pgdata:/var/lib/postgresql/data`) is that data was being mounted at `/var/lib/postgresql` — the parent — so it wasn't actually persisted at the path the user asked for.

## Fix

`container` 1.0.0 supports named volumes natively and auto-creates them on first `run`. This maps a Compose named volume straight to a `container` named volume named `<project>_<source>` (mirroring Docker Compose's namespacing), mounted at the **declared destination**.

Native volumes are block devices the guest owns, so `chown` works, and the destination is no longer rewritten to its parent. This also retires the now-outdated warning text ("the `container` tool does not support named volume references"), which predates 1.0.0.

## Verification

Against a real `docker-compose.yml` with a `redis` service using `redisdata:/data`:

- **Before:** container crashes on startup — `failed to find target executable docker-entrypoint.sh`.
- **After:** `backend_redisdata` volume is auto-created and redis reaches `Ready to accept connections` / `redis-cli ping` → `PONG`.

## Tests

Adds coverage to `ComposeVolumeTests` (the named-volume branch previously had none):
- maps to a native, project-namespaced volume
- #96 regression: a single-segment destination must not collapse to `/`
- nested destination + mode is preserved
- missing project name is skipped

All 137 static tests pass. (The `.containerDependent` dynamic suites fail in my environment with a pre-existing `No such file or directory` at setup — reproduced identically on a clean `main`, unrelated to this change.)

## Note / out of scope

A postgres service using a named volume will still hit the postgres image's `initdb: directory ... not empty ... lost+found` guard, because `container`'s native volumes are formatted filesystems with a `lost+found`. That's image/platform-specific and is worked around user-side by pointing `PGDATA` at a subdirectory of the mount; it's independent of this fix.

### DIFF
--- a/Sources/Container-Compose/Helper Functions.swift
+++ b/Sources/Container-Compose/Helper Functions.swift
@@ -198,19 +198,28 @@ func composeVolumeToRunArgs(
         }
     } else {
         guard let projectName else { return [] }
-        let volumeUrl = URL.homeDirectory.appending(path: ".containers/Volumes/\(projectName)/\(source)")
-        let volumePath = volumeUrl.path(percentEncoded: false)
-        let destinationUrl = URL(fileURLWithPath: destination).deletingLastPathComponent()
-        let destinationPath = destinationUrl.path(percentEncoded: false)
 
-        print(
-            "Warning: Volume source '\(source)' appears to be a named volume reference. The 'container' tool does not support named volume references in 'container run -v' command. Linking to \(volumePath) instead."
-        )
-        try fileManager.createDirectory(atPath: volumePath, withIntermediateDirectories: true)
+        // Named volume. As of `container` 1.0.0 these are supported natively
+        // (`container volume`) and are auto-created on first `run`, so we map a
+        // Compose named volume straight to a `container` named volume.
+        //
+        // We deliberately do NOT fall back to bind-mounting a host directory:
+        //   1. Bind mounts use virtiofs, whose mount the guest cannot `chown`.
+        //      Images whose entrypoint chowns its data dir (postgres, redis, …)
+        //      then die with "chown: Operation not permitted". Native volumes are
+        //      block devices the guest owns, so chown works.
+        //   2. The previous host-dir mapping stripped the destination to its parent
+        //      (`destination.deletingLastPathComponent()`), so `redisdata:/data`
+        //      mounted over `/` and shadowed the image's root filesystem (hiding its
+        //      entrypoint). See issue #96.
+        //
+        // The volume is namespaced as `<project>_<source>`, mirroring Docker
+        // Compose, so stacks don't collide on shared volume names.
+        let volumeName = "\(projectName)_\(source)"
 
         args.append("-v")
         let modeStr = mode.map { ":\($0)" } ?? ""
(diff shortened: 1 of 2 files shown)


Cyb3rDudu (commented):
Nice - this is a real fix. The `destination.deletingLastPathComponent()` collapsing `/data` down to `/` and shadowing the entrypoint is exactly #96, and moving to native volumes + the declared destination is the right call. The tests covering the regression and the nil-project case are great too.

Heads-up on an overlap: #119 rewrites this same branch with a more complete take — it honors the top-level `volumes:` block (`name` overrides, `external: true` skip-create, `driver` / `driver_opts` / `labels`) and replaces the `createVolumeHardLink` loop with a native `container volume create`. As written here those get dropped, and `createVolumeHardLink` in ComposeUp would still run, creating host dirs that nothing mounts anymore.

What this PR gets right that #119 doesn't: the `<project>_<source>` namespacing. #119 names the volume just `data`, so two projects would collide.

My plan is to take #119's handling as the base and fold your `<project>_<source>` default into it (I've left that suggestion on #119) - one coherent design instead of two. So I'm going to hold this one rather than merge it as-is, but the namespacing idea is yours and I'll credit it when I wire it in. Really appreciate you digging into #96; happy to ping you when the combined version is up so you can sanity-check it.

Cyb3rDudu:
closing — #119 merged with your <project>_<source> namespacing folded in, so this is covered. thanks for the idea!

PR #125, opened 2026-07-01
fix(service): add mem_limit and extra_hosts support

ryan106:
## Problem

Two commonly-used Compose keys are silently ignored by `container-compose`:

- **`mem_limit`** — the top-level memory limit shorthand (`mem_limit: 2g`). Services with high memory requirements (e.g. litellm, which spikes above 1 GB on startup) start without any limit and may be OOM-killed.
- **`extra_hosts`** — additional `/etc/hosts` entries (`extra_hosts: ["logto.localhost:host-gateway"]`). Without this, services that need to reach a host-side reverse proxy by hostname can't resolve it inside the container.

### `mem_limit`

- Added `public let mem_limit: String?` to `Service`
- Decoded as `String`; also handles bare integer bytes (YAML `mem_limit: 536870912`) via an `Int` fallback
- Mapped to `--memory` in `ComposeUp`, taking precedence over `deploy.resources.limits.memory` when both are present

### `extra_hosts`

- Added `public let extra_hosts: [String]?` to `Service`
- Accepts both Compose spec forms: list (`["hostname:IP"]`) and map (`{hostname: IP}`)
- The special token `host-gateway` is resolved to the host machine's default gateway IP via `route -n get default` (lazy — only runs when at least one entry uses the token)
- Passed to `container run` as `--add-host hostname:IP` (colon separator, per OCI standard)

## Changes

- `Sources/Container-Compose/Codable Structs/Service.swift` — new properties, `CodingKeys`, memberwise init params, and decoder branches
- `Sources/Container-Compose/Commands/ComposeUp.swift` — `--memory` flag from `mem_limit`; `--add-host` loop with lazy `host-gateway` resolution; `resolveHostGatewayIP()` helper
- `Tests/Container-Compose-StaticTests/MemLimitExtraHostsTests.swift` — 10 new tests covering string/integer/absent `mem_limit`, list/map/host-gateway/absent/multiple `extra_hosts`, and `resolveHostGatewayIP` return type

## Tests

```
swift test --filter MemLimitExtraHostsTests
```

All 10 tests pass; existing tests unaffected.

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Service.swift
+++ b/Sources/Container-Compose/Codable Structs/Service.swift
@@ -104,13 +104,23 @@ public struct Service: Codable, Hashable {
     /// Allocate a pseudo-TTY (-t flag for `container run`)
     public let tty: Bool?
     
+    /// Memory limit shorthand (e.g., "512m", "1g") — top-level alternative to
+    /// `deploy.resources.limits.memory`. Takes precedence when both are set.
+    public let mem_limit: String?
+
+    /// Additional `/etc/hosts` entries injected into the container. Each entry is a
+    /// `"hostname:IP"` string. The special token `host-gateway` resolves to the host
+    /// machine's IP as seen from inside the container.
+    public let extra_hosts: [String]?
+
     /// Other services that depend on this service
     public var dependedBy: [String] = []
-    
+
     // Defines custom coding keys to map YAML keys to Swift properties
     enum CodingKeys: String, CodingKey {
         case image, build, deploy, restart, healthcheck, volumes, environment, env_file, ports, command, depends_on, user,
-             container_name, labels, networks, hostname, entrypoint, privileged, read_only, working_dir, configs, secrets, stdin_open, tty, platform
+             container_name, labels, networks, hostname, entrypoint, privileged, read_only, working_dir, configs, secrets, stdin_open, tty, platform,
+             mem_limit, extra_hosts
     }
     
     /// Public memberwise initializer for testing
@@ -140,6 +150,8 @@ public struct Service: Codable, Hashable {
         secrets: [ServiceSecret]? = nil,
         stdin_open: Bool? = nil,
         tty: Bool? = nil,
+        mem_limit: String? = nil,
+        extra_hosts: [String]? = nil,
         dependedBy: [String] = []
     ) {
         self.image = image
@@ -167,6 +179,8 @@ public struct Service: Codable, Hashable {
         self.secrets = secrets
         self.stdin_open = stdin_open
         self.tty = tty
+        self.mem_limit = mem_limit
+        self.extra_hosts = extra_hosts
         self.dependedBy = dependedBy
     }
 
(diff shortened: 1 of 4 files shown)


Cyb3rDudu (changes requested):
@ryan106 The parsing and the `mem_limit` → `--memory` work look good, and the
tests are solid. One thing I'd like to nail down before merging: the
`host-gateway` resolution in `resolveHostGatewayIP`.

`route -n get default` returns the macOS host's **default-route gateway** -
i.e. the LAN router - not the host itself. On my machine that's `192.168.1.1`
while the host is `192.168.1.254`, so
`extra_hosts: ["logto.localhost:host-gateway"]` would resolve
`logto.localhost` to the *router*, not the host-side proxy it's meant to
reach. Docker's `host-gateway` means "the host as seen from the container
network," and the Mac's outbound default route isn't that — under Apple
`container`'s VM networking the container reaches the host via the VM bridge,
which `route -n get default` doesn't describe.

A couple of ways forward:
- Smoke-test it: `container run` with the `--add-host`, then check from inside
  the container whether the resolved IP actually reaches the macOS host.
- Or check whether Apple `container run` accepts `host-gateway` literally /
  exposes the correct gateway directly - if so, passing the token through is
  more robust than parsing `route`.

Minor, related: `needsGateway` checks the raw (pre-`resolveVariable`) entries,
so a variable wrapping the whole entry (`"${HOST_ENTRY}"` → `"foo:host-gateway"`)
would leave `hostGatewayIP` empty and emit `--add-host foo:`. Checking the
resolved value instead would close that.

Happy to merge once the host-gateway value is verified (or switched to whatever
Apple `container` natively supports).

ryan106:
> @ryan106 The parsing and the `mem_limit` → `--memory` work look good, and the tests are solid. One thing I'd like to nail down before merging: the `host-gateway` resolution in `resolveHostGatewayIP`.
> 
> `route -n get default` returns the macOS host's **default-route gateway** - i.e. the LAN router - not the host itself. On my machine that's `192.168.1.1` while the host is `192.168.1.254`, so `extra_hosts: ["logto.localhost:host-gateway"]` would resolve `logto.localhost` to the _router_, not the host-side proxy it's meant to reach. Docker's `host-gateway` means "the host as seen from the container network," and the Mac's outbound default route isn't that — under Apple `container`'s VM networking the container reaches the host via the VM bridge, which `route -n get default` doesn't describe.
> 
> A couple of ways forward:
> 
> * Smoke-test it: `container run` with the `--add-host`, then check from inside
>   the container whether the resolved IP actually reaches the macOS host.
> * Or check whether Apple `container run` accepts `host-gateway` literally /
>   exposes the correct gateway directly - if so, passing the token through is
>   more robust than parsing `route`.
> 
> Minor, related: `needsGateway` checks the raw (pre-`resolveVariable`) entries, so a variable wrapping the whole entry (`"${HOST_ENTRY}"` → `"foo:host-gateway"`) would leave `hostGatewayIP` empty and emit `--add-host foo:`. Checking the resolved value instead would close that.
> 
> Happy to merge once the host-gateway value is verified (or switched to whatever Apple `container` natively supports).

Awesome thanks for the comments! Changes pushed. Thanks Again!

Cyb3rDudu (changes requested):
We are getting there.

Bind-mounting over /etc/hosts replaces the container's entire hosts file, including the entry Apple container normally generates for the container's own hostname/IP. The generated file only has localhost + the extra_hosts lines, so anything inside the container that resolves its own hostname loses that mapping (Docker's --add-host appends; this overwrites). For the PR's use case (reaching a host proxy) it's fine, and there's no clean way to read the runtime-generated /etc/hosts before container run to merge — so it's somewhat inherent to the workaround. 
Can you
- add a one-line print warning when extra_hosts is used (e.g. "Note: extra_hosts replaces the container's /etc/hosts; the container's own hostname entry is overridden"), and/or
- include the container's own hostname in the generated file (e.g. 127.0.0.1 <containerName>) so self-resolution still works.

Minor

- The temp hosts file (NSTemporaryDirectory()...) is never cleaned up. One per service per up. Negligible (OS temp), but trivial to remove after the run if timing allows.
- Gateway uses service.networks?.first ?? "default"; for a service on multiple custom networks the first network's gateway may not be the host-reachable one. Edge case.

Cyb3rDudu (approved):
@ryan106 thanks, a pleasure to work with you

PR #127, opened 2026-07-02
Fix Up Command Foreground Interactions when sending signals

adrum:
When hitting ctrl+c while running the up command in the foreground, I noticed unexpected (to me) behavior. I was expecting it to behave like docker compose.

This allows for proper signal listening. SIGINT (first ctrl+c) gracefully kills the stack. SIGKILL (second ctrl+c) force kills the stack.


Cyb3rDudu (changes requested):
nice one, the dispatchsource signal handling is the right way around the
`SIG_IGN` that containerapiservice leaves set, and the two-tap graceful→force
+ auto-exit-when-stopped monitor are great.

one change before merging: `runForegroundUntilStopped` rebuilds names as
`<project>-<service>`, so it misses explicit `container_name` and the dotted
`<service>.<dnsDomain>` names from #97. for those, `stopContainers` /
`killContainers` / the monitor do `client.get(id: "<project>-<service>")` →
nil → skip, so ctrl-c exits container-compose but leaves the containers
running. pass the resolved names instead of service keys —
`services.map { containerName(for: $0.serviceName) }` — or call
`containerName(for:)` inside.

minor (non-blocking): the monitor needs to see each container running once
before it'll return, so a fast one-shot that exits before the monitor starts
never satisfies it. the signal path still exits fine.

Cyb3rDudu:
#126 modifies WaitForeverCpuTests.swift (the baseline fix); #127 deletes it (replacing waitForever with runForegroundUntilStopped). You need a small rebase on that file.

adrum:
I added support for the pattern defined in #97 for container name resolution, and I also backported the #126 idle-baseline technique into ForegroundWaitCpuTests.

Let me know if you see anything else!

Cyb3rDudu (approved):

PR #129, opened 2026-07-10
fix: tolerate env-var placeholders and quoted scalars in deploy.replicas decoding

### LINKED ISSUE #95: Got error "Expected to decode Int but found Scalar instead."  if i use an env var in an assignment to  "replicas"
Hi,

stumbled upon this:
```
Error: DecodingError.typeMismatch: expected value of type Int. Path: services.<oo.deploy.replicas. Debug description: Expected to decode Int but found Scalar instead.
```
when running 
```
container-compose up --env-file .env -f compose.yml
```

Relevant parts of the Docker compose yaml looks like this

```
    deploy:
      mode: replicated
      replicas: ${CORES}
``` 

and in `.env` is
```
CORES=1
```
(not `CORES="1"`)

I assume the Decoder looks at this before the variable is resolved...? 
So it seed the string "${CORES}" instead of the "1", as the value, from the .env file?

Can i work around this somehow? (other than hard coding the values in the compose file...)

mvanhorn:
## Summary
Give `Deploy` a custom `init(from:)` following the lenient-decoding precedent already established in `Service.init(from:)` (Service.swift, which uses try-Int/try-String fallbacks for `mem_limit`, `command`, `entrypoint`, etc.). Add an explicit `CodingKeys` enum for `mode`, `replicas`, `resources`, `restart_policy`. Decode `replicas` by first attempting `Int`; on failure, decode the `String` scalar, trim whitespace and surrounding quotes, and parse it with `Int(_:)` so quoted numerics like `"2"` still yield an integer.

## Why this matters
Running `container-compose up --env-file .env -f compose.yml` fails with `DecodingError.typeMismatch: expected value of type Int ... Expected to decode Int but found Scalar instead` when a compose file uses an environment variable for `deploy.replicas` (e.g. `replicas: ${CORES}` with `CORES=1` in `.env`). The root cause is decode ordering: `ComposeUp` decodes the YAML into `DockerCompose` (ComposeUp.swift line 164) before the env file is even loaded (line 167), and env-var interpolation happens per-field via `resolveVariable` after decoding.

See https://github.com/Mcrich23/Container-Compose/issues/95.

## Testing
- Happy path: a compose file with `deploy: {mode: replicated, replicas: 3}` still decodes with `replicas == 3` (no regression on plain Int scalars). - Quoted numeric: `replicas: "2"` decodes with `replicas == 2` instead of throwing typeMismatch. - Exact issue repro: the snippet from issue #95 (`mode: replicated`, `replicas: ${CORES}`) decodes successfully; `replicas` is nil and `replicasExpression == "${CORES}"`, and sibling fields (`mode`, `resources`) are unaffected. - Error path: a structurally invalid `replicas` value that is neither Int nor String scalar (e.g.

Fixes #95

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Deploy.swift
+++ b/Sources/Container-Compose/Codable Structs/Deploy.swift
@@ -21,6 +21,8 @@
 //  Created by Morris Richman on 6/17/25.
 //
 
+import Foundation
+
 
 /// Represents the `deploy` configuration for a service (primarily for Swarm orchestration).
 public struct Deploy: Codable, Hashable {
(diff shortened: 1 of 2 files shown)


Cyb3rDudu (approved):

mvanhorn:
Appreciate the merge, @Cyb3rDudu - env-var placeholders and quoted scalars in deploy.replicas decode tolerantly now.

PR #139, opened 2026-07-20
feat: error on unknown service name in `up` instead of hanging

OrtegaMatias:
## Summary

`container-compose up <service>` silently selected **nothing** when `<service>` wasn't defined in the compose file (a typo, or a service that was renamed/removed). In foreground mode (`up` without `-d`), it then blocked forever in `runForegroundUntilStopped`, waiting on containers that were never created — so a mistyped service name looked like a hang.

This validates requested service names up front and fails fast with `no such service: <name>`, matching `docker compose up <svc>`.

## Reproduce (before)

```console
$ container-compose up nginx-typo      # nginx-typo not in compose
# …hangs indefinitely, no output, has to be Ctrl-C'd
```

## After

```console
$ container-compose up nginx-typo
Error: no such service: nginx-typo
```

Verified end-to-end against a real ~30-service compose file: the unknown name now errors immediately, and a valid service name proceeds unchanged.

## Changes

- Add `ComposeError.noSuchService(String)` → `"no such service: <name>"`.
- Add `Service.validateRequestedServices(_:against:)` and call it in `ComposeUp.run()` right after the service list is built (before profile selection).
- Add unit tests in `Container-Compose-StaticTests` covering: all-exist passes, empty request passes, unknown throws, the error wording, and first-unknown-wins.

## Tests

`swift test --filter Container_Compose_StaticTests` → **215 tests, 20 suites, 0 failures** (same command the CI runs).

No behavior change for valid service names; purely additive validation.

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Service.swift
+++ b/Sources/Container-Compose/Codable Structs/Service.swift
@@ -422,6 +422,26 @@ public struct Service: Codable, Hashable {
         return sorted
     }
 
+    /// Validates that every explicitly requested service name exists among the
+    /// services defined in the compose file. Mirrors `docker compose up <svc>`,
+    /// which fails with "no such service: <svc>" rather than silently selecting
+    /// nothing — the latter previously left a foreground `up` hanging forever
+    /// (see `ComposeUp.runForegroundUntilStopped`).
+    /// - Parameters:
+    ///   - requested: Service names passed on the command line (may be empty).
+    ///   - defined: All services declared in the compose file.
+    /// - Throws: `ComposeError.noSuchService` for the first requested name that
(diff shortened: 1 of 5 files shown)


Cyb3rDudu:
Rebased onto latest main (includes #129 + #138) and pushed to the PR branch.

Changes vs original PR:
- Validation now runs after projectOptions.resolve() since topo-sort/select moved into the shared layer (#138)
- Validates against all compose services (pre-profile filtering) so explicit profile-gated names still work
- Removed redundant print warning in ComposeProject.orderedServices() since this hard error supersedes it

All 236 static tests pass (0 failures).

Cyb3rDudu (approved):

PR #141, opened 2026-07-28
Add ps subcommand to list project containers

adrum:
Lists the compose project's containers in an aligned NAME/IMAGE/STATUS/PORTS table, showing only running containers unless -a/--all is passed. Containers are matched primarily by the com.docker.compose.project/service labels stamped since #126 (reliable across naming modes), falling back to the candidate container names for containers created before the labels existed.

Resolves the project through the shared ComposeProjectOptions (#138) and fails fast on unknown service names with the noSuchService contract #139 established for up.

### DIFF
--- a/Sources/Container-Compose/Application.swift
+++ b/Sources/Container-Compose/Application.swift
@@ -31,6 +31,7 @@ public struct Main: AsyncParsableCommand {
             ComposeUp.self,
             ComposeDown.self,
             ComposeBuild.self,
+            ComposePs.self,
             Version.self
         ])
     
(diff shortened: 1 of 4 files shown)


Mcrich23:
Sorry for the slow response! I wasn't able to work on open source for most of the summer, but I'm catching up now. This is a useful addition.

One issue: the name fallback can include a container whose labels identify it as belonging to another project. Can you limit that fallback to containers without project labels? Please add coverage for project filtering, service filtering, and stopped containers.

adrum:
I added the fallback to limit to those without compose labels. Also added tests for project filtering, service filtering, and stopped containers.

I also noticed up/down's pre-start/stop cleanup has the same issue in regards to labels. We can address that in a future PR.

PR #142, opened 2026-07-28
Add pull subcommand to fetch service images from a compose file

adrum:
Pulls each resolved service's image, skipping build-only services (no image key) and images already present locally — mirroring the already-present match logic up uses (exact, registry-prefixed, or short reference). Passes the service's platform: through when set.

Resolves the project through the shared ComposeProjectOptions (#138) and fails fast on unknown service names per the #139 contract.

### DIFF
--- a/Sources/Container-Compose/Commands/ComposePull.swift
+++ b/Sources/Container-Compose/Commands/ComposePull.swift
@@ -1,0 +1,92 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Morris Richman and the Container-Compose project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Mcrich23:
Thank you for adding this! Skipping images that already exist means `pull` can't update something like `latest`. Can you fix that and resolve variables in image names before pulling? Please also check service selection: we shouldn't automatically pull dependencies unless requested. A test of the pull behavior would be helpful.

adrum:
Fixed all three: 
- pull now always pulls (the skip was up's semantics, and it stays there)
- image refs resolve `${VAR}/${VAR:-default}` from the env file first
- explicit `pull <service> no longer drags in `depends_on`. Added --include-deps to opt in to match docker compose

Also added united tests for each behavior.

PR #143, opened 2026-07-28
feat: support build.args list form and add `build --build-arg` flag

### LINKED ISSUE #69: build command: add support for --build-arg flag
## Summary
The build subcommand does not currently support the `--build-arg` flag, which is a standard part of docker compose build. Any project that relies on build-time arguments (e.g. for injecting private registry tokens or license keys into a multi-stage `Dockerfile`) cannot use `container-compose` build as a drop-in replacement today.

## Current Behavior
Running the following command fails immediately with an unrecognized option error:
```sh
container-compose build \
  --build-arg BO_NPM_PKG_AUTH_TOKEN=$BO_NPM_PKG_AUTH_TOKEN \
  --build-arg FONTAWESOME_PRO_TOKEN=$FONTAWESOME_PRO_TOKEN \
  --build-arg SIDEKIQ_KEY=$SIDEKIQ_KEY
Error: Unknown option '--build-arg'
Usage: container-compose <subcommand>
  See 'container-compose --help' for more information.
```

## Expected Behavior
Parity with docker compose build, which accepts one or more `--build-arg KEY=VALUE` pairs and forwards them to the underlying image build process as build-time variables. The equivalent native container build CLI also supports `--build-arg`, so this should be a straightforward passthrough.
```sh
# Should work the same as it does with docker compose
container-compose build \
  --build-arg BO_NPM_PKG_AUTH_TOKEN=$BO_NPM_PKG_AUTH_TOKEN \
  --build-arg FONTAWESOME_PRO_TOKEN=$FONTAWESOME_PRO_TOKEN \
  --build-arg SIDEKIQ_KEY=$SIDEKIQ_KEY
```

## Motivation / Use Case
Build args are essential for projects that need to inject secrets or configuration at image build time without baking them into the image or committing them to source control. Common real-world examples include:

* Private npm registry auth tokens (e.g. GitHub Packages, Artifactory)
* Font Awesome Pro license keys
* Private gem/package credentials (e.g. Sidekiq Pro)
* Any multi-stage Dockerfile that uses ARG to gate access to private dependencies

Without --build-arg support, teams adopting Apple Container via container-compose are blocked from migrating existing Docker-based CI/CD workflows that rely on this flag.

## Proposed Implementation
The build command should be updated to:

Accept --build-arg as a repeatable option (i.e. it can be specified multiple times), taking a KEY=VALUE string each time.
For each service being built, pass each provided --build-arg value through to the underlying container build invocation via its own --build-arg flags.

## References

[docker compose build docs — --build-arg](https://docs.docker.com/reference/cli/docker/compose/build/)
[container build CLI](https://github.com/apple/container) — already supports --build-arg

I appreciate all the work you've done on this project. 🍻

### LINKED ISSUE #136: DecodingError.typeMismatch in build args
Hi,

I got this error when starting my compose stack

```
Error: DecodingError.typeMismatch: expected value of type Mapping. Path: services.nginx.build.args. Debug description: Expected to decode Mapping but found Node instead.
```
when running 

```
container-compose up
```

Relevant parts of the Docker compose yaml looks like this:

```
nginx:
    container_name: proxy
    stop_grace_period: 1us
    build:
      context: nginx
      args:
        - DIR=development
```

OrtegaMatias:
## Summary

Completes `build.args` support end-to-end:

- **Fixes #136** — `build.args` in **list form** (`- KEY=VALUE`) threw `DecodingError.typeMismatch: expected Mapping ... found Node` and crashed the entire `up`. `Build` now decodes **both** the map form and the list form.
- **Closes #69** — adds a repeatable **`--build-arg KEY=VALUE`** flag to the `build` subcommand, matching `docker compose build`.

### #136 — list-form `build.args`
`Build.args` only decoded `[String: String]` (map form), so this valid Compose file crashed at decode time:

```yaml
nginx:
  build:
    context: nginx
    args:
      - DIR=development        # ← list form
```

`Build` now accepts map **or** list, reusing the exact same list-parsing semantics `environment:` already uses. To avoid duplicating that logic, the body of `Service.parseEnvironmentList` is lifted into a shared `parseComposeKeyValueList(_:)` helper that both `environment:` and `build.args:` call — split on the **first** `=` (so values keep internal `=`), and a bare `KEY` inherits from the environment.

### #69 — `build --build-arg`
```
container-compose build --build-arg TOKEN=$TOKEN --build-arg ENV=prod
```
Previously failed with `Unknown option '--build-arg'`. Now, per Docker parity:
- compose-file `args:` values are interpolated against the environment (unchanged),
- CLI `--build-arg` values are passed through as-is and **win on key conflicts**,
- a bare `--build-arg KEY` inherits its value from the environment.

The merge is a small pure function `mergedBuildArgs(fileArgs:cliArgs:)`, unit-tested independently of the daemon.

## Testing (TDD)

Written test-first. Full static suite is green locally:

```
✔ Test run with 243 tests in 21 suites passed
```

New static tests:
- list-form `build.args` decode + the exact #136 repro at the `DockerCompose` level
- list-form value with an internal `=` (DSN) round-trips
- `--build-arg` flag parsing (repeated + default-empty)
- CLI-overrides-file precedence, and non-colliding CLI+file args both kept

No changes to the dynamic (daemon) tests.

## Scope

`--build-arg` is added only to the `build` subcommand (parity with Docker). `environment:` behavior is unchanged (its existing list-form tests still pass after the shared-helper extraction).

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Build.swift
+++ b/Sources/Container-Compose/Codable Structs/Build.swift
@@ -42,7 +42,17 @@ public struct Build: Codable, Hashable {
             let keyedContainer = try decoder.container(keyedBy: CodingKeys.self)
             self.context = try keyedContainer.decode(String.self, forKey: .context)
             self.dockerfile = try keyedContainer.decodeIfPresent(String.self, forKey: .dockerfile)
-            self.args = try keyedContainer.decodeIfPresent([String: String].self, forKey: .args)
+            // `args:` accepts both forms per the Compose spec:
+            //   args:                 args:
+            //     KEY: value            - KEY=value
+            // The list form previously threw `typeMismatch` and crashed `up` (#136).
+            if let asMap = try? keyedContainer.decodeIfPresent([String: String].self, forKey: .args) {
+                self.args = asMap
+            } else if let asList = try? keyedContainer.decodeIfPresent([String].self, forKey: .args) {
+                self.args = parseComposeKeyValueList(asList)
+            } else {
+                self.args = nil
+            }
         }
     }
 
--- a/Sources/Container-Compose/Codable Structs/Service.swift
+++ b/Sources/Container-Compose/Codable Structs/Service.swift
@@ -369,17 +369,7 @@ public struct Service: Codable, Hashable {
     ///                                   the host doesn't define it, falls
     ///                                   back to an empty string)
     static func parseEnvironmentList(_ entries: [String]) -> [String: String] {
-        var dict: [String: String] = [:]
-        for entry in entries {
-            if let eqIdx = entry.firstIndex(of: "=") {
-                let key = String(entry[..<eqIdx])
-                let value = String(entry[entry.index(after: eqIdx)...])
-                dict[key] = value
-            } else {
-                dict[entry] = ProcessInfo.processInfo.environment[entry] ?? ""
-            }
-        }
-        return dict
+        parseComposeKeyValueList(entries)
     }
 
     /// Returns the services in topological order based on `depends_on` relationships.
--- a/Sources/Container-Compose/Commands/ComposeBuild.swift
+++ b/Sources/Container-Compose/Commands/ComposeBuild.swift
@@ -45,6 +45,12 @@ public struct ComposeBuild: AsyncParsableCommand, @unchecked Sendable {
     @Flag(name: .long, help: "Do not use cache when building")
     var noCache: Bool = false
 
+    @Option(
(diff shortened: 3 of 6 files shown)


Mcrich23:
I like this, especially since it addresses both #69 and #136. Can you make invalid argument formats throw instead of silently dropping them? Please also check bare arguments with no environment value so we don't accidentally replace a Dockerfile default with an empty string. A test confirming the arguments reach the builder would be helpful. The shared helper can stay internal.

Issue #128, opened 2026-07-09
Containers launch sequentially - can they be launched concurrently?


swwol:
Finding that using container-compose with a `docker-compose.yml` file containing 11 wiremock servers, each servers is launched  sequentially, waiting for each to complete before starting to launch the next. This means the whole launch is pretty slow for multiple containers. 

This is exacerbated by a timeout issue on 1.0.0 tagged version, but not on main, the launch time for 11 services is over 5 minutes as each waits for 30 seconds. But even on main the sequential launching takes ~23 seconds. 

When launching the services with a custom bash script in parallel, this is reduced to 7 seconds. 

Is it feasible to launch the services in parallel?

Mcrich23:
Sorry for the slow response! I'm catching up on issues after being unable to work on open source for most of the summer.

Yes, independent services should be able to start concurrently. #144 and #145 work toward that, with #154 addressing a dependency issue it exposed. I want to verify dependency handling and failures before making it the default. Your 11-service example would be helpful for comparing startup time afterward if you don't mind providing some numbers and testing afterwards.

PR #144, opened 2026-07-28
feat: add dependencyLevels scheduler for parallel `up` (#128, step 1)

OrtegaMatias:
## Summary

First step toward parallel `up` (#128): a pure, daemon-free scheduler that groups services into launch **waves** by their `depends_on` graph. Every service in a wave has all of its in-set dependencies in an earlier wave, so a whole wave can be `container run` launched concurrently.

Today `up` launches services strictly sequentially (`for target in project.services { try await configService(...) }`), so N independent services pay N× the per-service start+readiness cost. The issue reports ~23s for 11 independent services vs ~7s launched in parallel.

This PR lands **only the schedulable core** — a pure function with tests — with **no behavior change** yet (`up` still launches serially). The concurrent launch that consumes it will follow in a separate PR, so the correctness-critical ordering logic can be reviewed and regression-tested on its own (and it lands in the only tier CI runs).

## API

```swift
static func dependencyLevels(
    _ services: [(serviceName: String, service: Service)]
) throws -> [[String]]
```

Mirrors the existing pure graph helpers (`topoSortConfiguredServices`, `selectServices`, `validateRequestedServices`):

- **Levels:** `level 0` = services with no in-set dependency; `level(s) = 1 + max(level(dep))`. Level 0 first; intra-wave order is unspecified (they start concurrently).
- **Out-of-set deps ignored:** a `depends_on` naming a service not in the set (e.g. profile-excluded) is treated as already satisfied — same tolerance as `topoSortConfiguredServices`.
- **Cycles throw before returning:** a dependency cycle (including a self-dependency) throws the same `NSError`/"ComposeError" shape as `topoSortConfiguredServices`, so a cyclic graph can never deadlock a future concurrent launch.

## Tests

New `@Suite("Parallel Scheduler Levels")`, static (no daemon), mirroring `ServiceDependencyTests`:

- linear chain `web → app → db` ⇒ `[["db"], ["app"], ["web"]]` (chain is **not** falsely parallelized)
- diamond (`front → {web, api} → db`) ⇒ siblings share a wave
- independent set ⇒ a single wave (the #128 case)
- dependency cycle ⇒ throws
- self-dependency ⇒ throws
- `depends_on` on an out-of-set service ⇒ dependent lands in wave 0, no throw

```
✔ Test run with 242 tests in 22 suites passed
```

### DIFF
--- a/Sources/Container-Compose/Codable Structs/Service.swift
+++ b/Sources/Container-Compose/Codable Structs/Service.swift
@@ -422,6 +422,56 @@ public struct Service: Codable, Hashable {
         return sorted
     }
 
+    /// Groups the selected services into launch WAVES for a parallel `up`. Every
+    /// service in `levels[k]` has all of its in-set `depends_on` dependencies in
+    /// some earlier wave, so every service in one wave can be `container run`
+    /// launched concurrently (issue #128). Level 0 (no in-set dependencies)
+    /// comes first.
+    ///
+    /// - `depends_on` entries naming a service that is not in `services` are
+    ///   ignored — mirroring `topoSortConfiguredServices` and the way
+    ///   profile-excluded dependencies are tolerated.
+    /// - Throws (same `NSError`/"ComposeError" shape as `topoSortConfiguredServices`)
+    ///   on a dependency cycle, including a self-dependency, *before* returning —
+    ///   so a cyclic graph can never deadlock the concurrent launch.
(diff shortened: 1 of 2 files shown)


Mcrich23:
This looks reasonable as the first step, and I appreciate the cycle tests. The static CI checks are passing. We should leave #128 open until the concurrent launch behavior is implemented and tested in #145.

Mcrich23 (approved):

Issue #140, opened 2026-07-23
Environment variables interpolation not supported


TheDogHusky:
Hey!

Just got my macbook pro, and decided to code my stuff on it with container compose.
The biggest issueI just came across is variable interpolation.

Here is the file:
```yaml
services:
    redis:
        image: redis:7-alpine
        container_name: victoria-redis
        ports:
            - "6379:6379"
        command: ["redis-server", "--requirepass", "${REDIS_PASSWORD:-devpassword}"]
        env_file:
            - .env
        volumes:
            - redis_data:/data
        restart: unless-stopped

volumes:
    redis_data:
```

and .env:
```
REDIS_PASSWORD=mypassword
```

You would expect to be able to connect to `redis://localhost:6379` with the password "mypassword" right?
Wrong! Unlike with docker, the password will be "${REDIS_PASSWORD:-devpassword} ", literally.

This has been a very annoying issue I came across.

Hoping it'll get fixed soon!

Cheers,
Adam

TheDogHusky:
Seems like [this pr](https://github.com/Mcrich23/Container-Compose/pull/102) fixes it

Mcrich23:
Sorry for the delay getting back to you. I wasn't able to work on open source for most of the summer. Command arguments currently aren't interpolated, which explains the literal password.

#146 addresses this specific case; #102 doesn't cover command arguments. Once #146 lands, would you mind trying your Redis example again and letting me know whether it uses the expected password?

TheDogHusky:
Of course, i'll be more than happy to try it out!

PR #146, opened 2026-08-07
fix: interpolate command and entrypoint environment

saeidakbari:
## Summary
- Resolve Compose variables in each list-form command and entrypoint token.
- Use the loaded project environment, preserving process-environment precedence.
- Add command and entrypoint regression coverage.

## Verification
- swift test --filter EntrypointCommandTests (16 tests passed)
- Rebuilt with swift build -c release
- Ran the release binary against the Artifex migration service; the prior literal-variable URL parse failure no longer occurs.


Mcrich23:
This looks like a good, focused fix for #140. Can you add a dynamic test confirming the interpolated value reaches the command? Please also check escaped dollar signs so values intended for the container's shell aren't expanded on the host.

